PR #6, opened 2026-08-13
Add authoritative noclip movement

Confetti3:
## Summary

Adds a practical exploration noclip mode to Sunrise by extending the existing teleport camera and physics hooks. No new signatures or detours are introduced.

While noclip is active, the camera hook polls the player's configured movement actions and accumulates a frame-rate-independent world displacement. The existing physics-sync hook applies that displacement to a stored rigid-body target, clears body velocity, and lets the original sync publish the overridden transform.

## Why

Writing the game-object or camera transform alone does not move the authoritative player body. Reading the body position and offsetting it each frame is also insufficient because gravity, collision depenetration, and ordinary movement may already have changed that position.

This change instead keeps an independent noclip target:

1. Initialize the target from the current local-player rigid-body position.
2. Accumulate input displacement into the target.
3. Write the target to the body immediately before the original physics sync.
4. Write zero velocity so gravity and prior momentum do not leak into the published transform.

That preserves the player entity, camera, streaming origin, and world membership while allowing the published position to override collision-resolved movement.

## Implementation

- Reuses the existing per-frame camera-transform and pre-publication physics-sync detours.
- Resolves forward, backward, left, right, jump, crouch, and sprint bindings once when noclip activates.
- Uses normalized camera forward, a planar right vector, and world Z for vertical movement.
- Normalizes combined input so diagonal movement is not faster.
- Applies configured units-per-second movement using `steady_clock` elapsed time.
- Clamps one frame to 50 ms to prevent an alt-tab or loading stall from producing a large jump.
- Keeps a stored body target rather than accumulating from physics output.
- Writes all body-velocity lanes to zero while active.
- Makes teleport requests update the shared noclip target so the next sync cannot undo them.
- Reuses the cached player component, forced-sync trampoline, and short synthetic-forward press for sleeping bodies.
- Forces a sync only after movement changed the target.

## Lifecycle and safety

Noclip target and pending movement state are invalidated when:

- noclip is disabled;
- the controlled-player handle becomes invalid;
- the cached component no longer owns the controlled player;
- the rigid-body pointer changes;
- camera/player state unloads; or
- the teleport hook shuts down.

Pending deltas retain the existing short request lifetime so movement entered during orbit or loading cannot be replayed against a later spawn. The change does not modify the body motion type, physics-component suppression byte, collision filters, or game data.

## Settings and interface

The existing Teleport page is shown as `Movement` and now contains a Noclip section with:

- feature availability;
- current active state;
- toggle key;
- movement speed; and
- sprint boost multiplier.

The values persist immediately in `teleport.json`. Missing noclip fields retain backward-compatible defaults, so existing teleport configurations continue to load.

Default values are 20 world units per second and a 4x sprint boost. A fresh configuration has noclip disabled and no toggle key, preventing accidental activation.

## Controls

- Forward/backward: existing movement bindings
- Strafe: existing left/right bindings
- Rise: jump binding
- Descend: hold-crouch binding, falling back to toggle-crouch
- Boost: hold-sprint binding, falling back to toggle-sprint
- Toggle: separately configured noclip key

Native movement inputs remain visible to the game in this first version. The authoritative target and velocity clear dominate their physical result, while those inputs also help keep the body active. Game-only input suppression can be added separately after broader runtime validation.

## Expected behavior and limitations

This is an authoritative-position exploration mode, not removal from the Havok world. It should allow traversal through static geometry and neutralize gravity, but trigger volumes, kill planes, scripted volumes, and destination/bubble transitions can still run. Disabling noclip inside solid geometry may cause ejection, snapping, or death, and very high speeds can outrun streaming.

## Validation

- `clang-format --dry-run --Werror` passed for all seven changed files using the repository configuration.
- `clang-tidy` completed for all five changed translation units using the repository configuration with no displayed user-code diagnostics.
- Release x64 clean rebuild passed with warnings treated as errors. The local machine has MSVC `v143`, so the build used a one-command `PlatformToolset=v143` override because the project requests unavailable `v145`; the project file was not changed.
- Manual testing on the target game build confirmed noclip activation and camera-relative movement. The initial strafe handedness was corrected, and the updated left/right direction was confirmed in game.

No compiled DLL, game data, runtime settings, or other generated artifacts are included in this PR.

### DIFF
--- a/Sunrise/src/client/hooks/teleport/runtime.h
+++ b/Sunrise/src/client/hooks/teleport/runtime.h
@@ -35,8 +35,11 @@ void uninstall() noexcept;
  */
 void capture_forward(std::uint32_t playerIndex) noexcept;
 
-/** Latches one teleport request if the bound key went down this frame. */
-void poll_request() noexcept;
+/** Polls the teleport request, noclip toggle, and active noclip movement controls. */
+void poll_controls() noexcept;
+
+/** @return True while noclip is toggled on and its feature setting remains available. */
+[[nodiscard]] bool noclip_active() noexcept;
 
 /** Runs the move for a request no physics tick collected, and drives the sync for it. */
 void force_pending() noexcept;
--- a/Sunrise/src/client/hooks/teleport/teleport_lifecycle.cpp
+++ b/Sunrise/src/client/hooks/teleport/teleport_lifecycle.cpp
@@ -67,7 +67,8 @@ template <typename T> [[nodiscard]] T original(std::size_t slot) noexcept {
 }
 
 /**
- * Publishes the forward vector and reads the bound key, then defers to the original.
+ * Publishes the forward vector and reads the movement controls, then defers to the original.
+ *
  * @param playerIndex Player whose camera was transformed.
  * @return Whatever the original returns.
  */
@@ -75,7 +76,7 @@ std::int64_t __fastcall camera_transform(std::uint32_t playerIndex) noexcept {
     const CameraTransform next = original<CameraTransform>(kCameraSlot);
     const std::int64_t result = next != nullptr ? next(playerIndex) : 0;
     capture_forward(playerIndex);
-    poll_request();
+    poll_controls();
     force_pending();
     return result;
 }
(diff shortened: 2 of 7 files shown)


Confetti3:
Stan noted that Havok might be a stronger approach with a simple hook over teleporting. I'm looking at a simpler solution.

PR #8, opened 2026-08-13
Add horizontal noclip

Confetti3:
## Summary

Adds horizontal noclip through one independent Havok-step hook. It resolves the character rigid body from the current simulation islands, advances an authoritative X/Y target with native velocity, and replaces the collision-resolved horizontal position after simulation. Havok continues to control vertical movement.

The Movement page adds persisted availability and toggle-key settings. Teleport only invalidates the noclip target after a successful move.

## Validation

- Tested movement through solid geometry and toggle behavior across death on the target build.
- Clean `Release|x64` build with the project's `v145` toolset.
- `clang-format --dry-run --Werror` and `git diff --check` pass.

## Scope

Vertical collision remains enabled for normal ground movement. Triggers, kill planes, and environmental damage are not disabled.

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -187,6 +187,7 @@
     <ClCompile Include="src\client\ui\runtime\client_ui_module_runtime.cpp" />
     <ClCompile Include="src\client\ui\teleport\teleport_panel.cpp" />
     <ClCompile Include="src\client\teleport\teleport_settings_store.cpp" />
+    <ClCompile Include="src\client\hooks\noclip\horizontal_noclip.cpp" />
     <ClCompile Include="src\client\hooks\teleport\teleport_lifecycle.cpp" />
     <ClCompile Include="src\client\hooks\teleport\teleport_move.cpp" />
     <ClCompile Include="src\client\hooks\teleport\teleport_action_key.cpp" />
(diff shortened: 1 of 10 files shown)


stanuwu:
Reviewed:
- horizontal_noclip.cpp:27-30 comment is inaccurate
- teleport_move.cpp doxygen and comments removed
- state between hooks is not thread safe, use atomic operations
- pointers should not be stored between hooks, it is unsafe and can cause unexpected issues since there is no guarantee the pointer stays valid
- install() fails both teleport and noclip if one pattern misses
- comments have random blank lines
- kMaximumStepSeconds is completely undocumented magic value
- g_target[kVerticalLane] written and never read
- g_target[0] and g_target[1] are not documented at all, magic numbers again
- clear_target, clear_player, reset_target, rebase -> all kinda do the same thing this is unneeded compelxity
- owns_player runs every tick which is not needed
- controlled_handle() called multiple times unnecessarily
- module should not be part of teleport, should be its own module
- definitions of header and code do no always line up like set_active, params are not correct
- before_simulation does unneeded casts
- horizontal_noclip.h does not document its parameters
- uninstall() order should not change without reason
- toolset is v145, thats what it should be built and tested on

Overall this change introduces a lot of state that is not needed. The general implementation is correct but complexity has to be reduced alongside fixing the other issues pointed out.

Confetti3:
Reworked this around the review: noclip is now an independent single Havok-step module, resolves the character body fresh from simulation islands, uses atomic target state, stores no gameplay pointers between calls, and installs separately from teleport. Tested through solid geometry and across death on 0.2.1; the v145 Release build and formatting checks pass.

stanuwu:
Thanks. Merged

PR #13, opened 2026-08-14
Fix client shutdown lock leak

AlexisSliwak:
This is a simple fix with a small scope, so it should be easy to approve.

Upon hooks::config_getter::uninstall() or hooks::assert_handler::uninstall() failing, runtime::g_lock is now released.
This ensures the runtime::g_lock is always released before returning from function that acquired this lock.
This lock is only used in 3 functions, each time being acquired at the start of the function and released by the time it returns. There is no reason why that pattern should be any different for these two return statements since the purpose of this lock is just to keep these functions from running concurrently.
Whether or not it is currently likely for someone to experience the theoretical freeze by this lock leak, we should get it out of the way.


stanuwu:
Good find, thanks

PR #14, opened 2026-08-14
feat: add fov changer 

blahajr:
## overview 

- added an fov tab to edit your fov - see attached video

https://github.com/user-attachments/assets/b2403af5-d2b0-4a12-bb51-09b2ce54f7d3

let me know any suggestions/changes

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -20,7 +20,7 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Debug'" Label="Configuration">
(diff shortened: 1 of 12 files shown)


stanuwu:
Seperate changes in seperate PR please.
FOV looks good, readme change will not be merged.

blahajr:
done :)

stanuwu:
Reviewed:
- please do not use raw RVA offsets, use patterns 
- please provide some information for the offset and the struct its in
- do not downgrade toolset, it should stay at v145
- fov is not persisted in any state, it will only stay as long as nothing else changes it, it will not stay through restarts
- hook is missing shutdown logic
- disable toggle does nothing since no original value was saved, this should restore the one from the settings
- use GetModuleHandleW(nullptr) instead of GetModuleHandleW(L"destiny2.exe")
- teleport_settings_store.cpp comment change not related
- please run lint and format
- fov_panel.cpp:605-609 wrong error and also smileys do not belong in code
- apply_fov failing does not log
- use uint16 for storage since its the same data type the game uses
- fov page missing a heading and will not look the same as other pages
- item placement in .vcxproj is incorrect

blahajr:
commit [50f4720](https://github.com/stanuwu/Sunrise/commit/50f4720171152d74f2d2639f20963445dcbedbb4) should address the problems you gave

stanuwu:
Sorry but this is not a good change.
Extra scanner just for this and polling thread? What are we doing?
Just use regular xref pattern from fov writes.

blahajr:
Closing PR for now as i dont have enough time to fix the issues for a week

PR #15, opened 2026-08-14
Use native assert-handler setter

jasonshaw0:
## Summary

- retain the native assert-handler setter selected alongside the voted handler slot
- use the game's setter for Sunrise handler installation and restoration
- preserve expected-value ownership checks before changing the slot

## Why

The resolver already identifies native assert-handler setter copies and derives the slot written by each copy, then votes those slots against sampled assert sites. Before this change, it kept the winning slot and original handler but discarded the setter address paired with that winning slot.

That meant the lifecycle wrote the slot directly despite the existing TODO to use the game's own setter. Retaining the setter selected by the same vote lets Sunrise follow the native mutation path without adding another signature or hard-coded address.

## Validation

- Release|x64 build succeeded with local-only `/p:PlatformToolset=v143` override
- `clang-format --dry-run --Werror` passed on changed files
- `clang-tidy` completed for changed translation units with no reported diagnostics
- `git diff --check` passed
- source audit confirmed no direct dereferenced assert-handler slot assignment remains in lifecycle

Runtime testing was not performed.

### DIFF
--- a/Sunrise/src/client/hooks/assert_handler/assert_handler_lifecycle.cpp
+++ b/Sunrise/src/client/hooks/assert_handler/assert_handler_lifecycle.cpp
@@ -11,21 +11,16 @@ bool g_installed{};
 
 namespace {
 
-/**
- * Exchanges the handler slot under its own page protection.
- * @param slot Resolved handler slot.
- * @param expected Value the slot must currently hold.
- * @param value Replacement pointer.
- * @return True when the slot holds the replacement after the call.
- */
-[[nodiscard]] bool exchange(std::byte** slot, const void* expected, void* value) noexcept {
-    if (slot == nullptr || *slot != expected) {
+/** Calls the native setter after the expected-value ownership check. */
+[[nodiscard]] bool set_handler(const targets::game::assert_handler::Targets& targets,
+                               const void* expected,
+                               void* value) {
+    if (targets.setter == nullptr || targets.slot == nullptr
+        || static_cast<const void*>(*targets.slot) != expected) {
         return false;
     }
-    // The slot is a writable data global, so the write needs no protection change. The game's own
-    // setter stores it with a plain move. TODO: call that setter instead of writing the slot.
-    *slot = static_cast<std::byte*>(value);
-    return true;
+    targets.setter(value);
+    return static_cast<void*>(*targets.slot) == value;
 }
 
 } // namespace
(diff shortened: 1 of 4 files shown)


stanuwu:
Reviewed:
- a fail will not log, just silently return false on uninstall (need to log failure here)
- winnerSetter == nullptr is dead check, we know this value is not null
- comments / docs have been removed, these should not be removed
- noexcept is not correct here, unless proven
- add setter at the end, not start -> keep field order consistent
- test on toolset v145

jasonshaw0:
Addressed in `90485cd`:

- uninstall now logs assert-handler restoration failure
- removed the dead `winnerSetter == nullptr` check
- restored the assert target docs
- removed the unproven `noexcept` from the native setter path
- moved `setter` to the end of `Targets`
- made the slot pointer comparisons explicit

Validation rerun:

- `Release|x64` build passed with local-only `/p:PlatformToolset=v143`
- `clang-format --dry-run --Werror` passed
- `clang-tidy` completed with no reported diagnostics
- `git diff --check` passed

I could not validate with `v145` locally because the installed VS Build Tools only include MSVC `14.44.35207`; the normal build fails before compile with `MSB8020` for missing `PlatformToolset=v145`.

stanuwu:
Please install v145 and make sure it builds, then also test ingame.
Can merge after that is confirmed

jasonshaw0:
Confirmed on PR head `90485cd`.

Validation:

- v145 `Release|x64` build passed using Visual Studio Build Tools 2026 / MSVC `14.50.35717`:
  `MSBuild.exe Sunrise\Sunrise.vcxproj /m /p:Configuration=Release /p:Platform=x64 /nologo`
- output DLL: `build\x64\Release\steam_api64.dll`
- deployed that DLL into the local Sunrise test install at `C:\Users\jj\Downloads\sunrise\bin\x64\steam_api64.dll`
- in-game smoke test passed: Destiny 2 launched successfully and reached character select
- runtime log confirms assert handler installed successfully: `client level=info ev=assert stage=install result=ok`

Ready to merge from my side. I attempted the merge, but GitHub rejected it because `jasonshaw0` does not have permission to execute `MergePullRequest`, so someone with repo merge access will need to merge it.

CrayonR6:
LGMT 👍 !

Voided24:
LGMT!

stanuwu:
Thanks

PR #17, opened 2026-08-14
Add CMake as a build system for Linux

TotalTaxAmount:
For far too long Destiny 2 has been unable to run on Linux. This mod allows that to finally happen (thank you!) with a few minor changes. In order to allow people to help develop the mod on a Linux based system this PR adds a CMake based build system. This allows for cross platform development. I have tested build this on both Windows and Linux and it does work (I did not test the dll on Windows as I do not have the old Destiny build on my Windows drive, though there is no reason why it won't work). 

# Build Instructions:
## Linux:
You need to have `clang`, `llvm`, and `xwin` installed. First download the windows headers with xwin:
```
xwin --accept-license splat --include-debug-libs --sdk-version 10.0.26100 --output .xwin-cache
```
Then run the build, make sure to include the linux-to-win-toolchain.cmake.
```
cmake -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE=$(pwd)/linux-to-win-toolchain.cmake
cmake --build build
```
## Windows
All I did was open the folder inside of Visual Studio and it recognized the CMakeLists.txt and I was able to just build it by clicking the green run button, though it complained about not having a executable target which makes sense. I have no experience with Visual Studio so please let me know if you need any changes.

Note: This PR does not include the changes required to make the mod run on Linux but I can PR those if this gets merged


stanuwu:
Hello, I am not comfortable with cmake as primary build system,
MSVC will stay the primary tooling but if you want you can PR an additional linux build script or support files to allow linux development.

TotalTaxAmount:
Ok, so you are ok with cmake aswell as MSVC?

stanuwu:
If its a small seperate build script that does not interfere with my MSVC tooling or need extra maintainance im fine with it.

TotalTaxAmount:
I brought back the MSVC files. I have also left the windows compatible stuff in the cmake files, it's  pretty small and I don't think it hurts leaving it there. (also my bad for all the pings I accidentally deleted my previous comment)

stanuwu:
Reviewed:
- require cmake version 3.31+ to ensure llvm-mt discovery
- embed liecense from vendor folder
- CMakeLists tracking is missing default setting json, could build with outdated file on partial build
- dont use recursive glob without CONFIGURE_DEPENDS
- pin xwin sdk version to ensure stable build

TotalTaxAmount:
Great thanks, all of those are done except pinning the xwin version. I can set a variable in CMakeLists for the version but right now the sdk if manually downloaded by the builder, I could add arguments to the build instructions or I could have cmake download the xwin headers if they are not there but that seems a little jank

stanuwu:
It would be good to pin a version just in case the build can fail on other versions. However its done, automatic download or the user having to install it themselves really doesn't matter.

TotalTaxAmount:
Done.

stanuwu:
Thanks! Will merge once I push my stashed changes.

stanuwu:
Thanks. Merged

PR #24, opened 2026-08-15
 Anchor signature scans on their first exact byte (~4s faster boot)

SkyzerFlyzer:
---
  Title: Anchor signature scans on their first exact byte (~4s faster boot)

  ---
  What

  Two commits:

  - b6ad2d9 — debug-level boot timing
  - 699c1cf — the scan change itself

  Why

  patterns::resolve_all and patterns::collect_matches tested every byte offset in every executable range. On the live game image that is ~105 MB across two .text sections.

  Activation doesn't sweep that once — it sweeps it repeatedly. There's the batched sweep in game_target_resolution.cpp, and then roughly 26 more: every scan_main_image_unique caller across bootflow, teleport, queuez, bitmap, banner, investment and the external-server
  guard re-inspects the PE and runs its own full pass for a single signature.

  The result was several seconds of boot spent scanning byte by byte, behind the initialization overlay. The overlay exists precisely because of this stall — activate_main_once wraps the sweep in busy::Task::initialization with the comment "The sweep stalls whichever
  thread calls it, so the overlay says what is happening."

  How

  Each pattern now takes an anchor — its first exact byte — and memchr jumps straight to the offsets where that byte lines up, skipping the gaps at memory speed. The full masked compare only runs at a candidate offset.

  is_valid folds into the new anchor_of. "Has at least one exact byte" was already the validity rule, and it is exactly what the anchor needs, so the two checks became one.

  No behavioural change. Matched addresses, first-match-wins selection, ambiguous-on-second-match, and the invalid / missing / unique statuses are all preserved. collect_matches still returns addresses in ascending order and still caps at the output size. The traversal
  order did change — it is now pattern-major and visits only anchor-hit offsets rather than every offset — but no result depends on traversal order.

  No game data added. The anchor is derived at compile time from signatures already in the tree; nothing new is embedded.

  Effects

  Measured on real boots, before and after, using the timing from the first commit.

  Before (naive sweep):
  core   level=debug ev=initialize phase=complete ms=31 result=ok
  client level=debug ev=activate stage=main phase=complete ms=7094 result=ok

  After (anchored sweep):
  core   level=debug ev=initialize phase=complete ms=32 result=ok
  client level=debug ev=activate stage=main phase=complete ms=3140 result=ok

  - ev=activate stage=main: 7,094 ms → 3,140 ms — ~4.0 seconds saved, a 2.26x speedup on that stage
  - ev=initialize: 31 ms → 32 ms (unchanged, as expected)

  ▎ Disclaimer. GetTickCount64 has ~15 ms granularity, so the ev=initialize figures sit within one tick of each other. They are a control showing the change is localised to activation, not a result. The ev=activate stage=main figure is the real one, and at ~4 s is far
  ▎ outside timer resolution or run-to-run variance. These are single runs on one machine — indicative of magnitude rather than precise.

  Both runs resolved every signature and attached all 39 hooks, with 11/11 boot-flow fixes installed and zero failures.

  The remaining ~3.1 s in that window is not scanning — it is hook installation and the rest of activation, previously hidden behind the sweep.

  About the timing commit

  It's included because it is what produced the numbers above, and it is how anyone diagnoses a future regression here — the activation window was previously one opaque block.

  It logs ms= on ev=initialize and ev=activate stage=main at debug level only, so a normal run at warn or info is completely unchanged. It follows the existing ms= / GetTickCount64 idiom already used by hash_name_build.cpp and spawn_set_build.cpp, down to computing a
  named const std::uint64_t elapsed and emitting ms= before result=.

  Three implementation notes:

  - Both boundaries use phase=begin / phase=complete rather than stage=. In the client those lines already carry stage=main, and in core the stage= field already enumerates real init stages (logging, ui, state, …) via report_stage_failure — overloading it with
  pseudo-stages would pollute that field.
  - The client timer starts after busy::begin, because begin can block up to 500 ms in wait_for_present on a cold boot. That wait belongs to the overlay, not to the work being measured.
  - The initialize begin marker sits immediately after log::initialize rather than at function entry, because that stage is what brings the sinks up — nothing logged before it can reach a channel. The reported duration still counts from function entry. The failure-path
  timing is emitted before the unwind so it measures initialization alone and stays comparable with the success line; a logging-stage failure has no sinks left and reports nothing.

  Verification

  - Built with the pinned toolchain — MSVC v145 (14.51.36231) and Windows SDK 10.0.26100: 479 files, 0 errors, 0 warnings with TreatWarningAsError enabled
  - clang-format clean on all touched files
  - Verified in game across multiple real boots

  Attribution

  Developed with Claude Code. Both commits carry Co-Authored-By trailers.

### DIFF
--- a/Sunrise/src/client/patterns/registry.cpp
+++ b/Sunrise/src/client/patterns/registry.cpp
@@ -1,23 +1,43 @@
 #include "registry.h"
 
+#include <cstring>
+
 namespace sunrise::client::patterns {
 namespace {
 
+/** Returned by next_candidate when a range holds no further anchor byte. */
+constexpr std::size_t kNoCandidate = static_cast<std::size_t>(-1);
+
 /**
- * Validates one pattern before the shared image sweep.
- * @param pattern Pattern name, bytes, and exact-byte mask.
- * @return True when the pattern has a name, bytes, and at least one exact byte.
+ * The one exact byte a pattern's candidate search keys on.
+ * Only exact bytes can be searched for, so a pattern without one cannot be scanned at all. That
+ * is the same condition the sweep already rejected patterns on, so the anchor doubles as the
+ * validity check.
  */
-[[nodiscard]] bool is_valid(const Pattern& pattern) noexcept {
+struct Anchor {
+    /** Position of the anchor byte inside the pattern. */
+    std::size_t index{};
+    /** The byte itself, held unsigned so it reaches memchr without sign extension. */
+    unsigned char value{};
+    /** False for a pattern the sweep cannot scan, which is the pattern it must reject. */
+    bool valid{};
+};
+
+/**
+ * Picks the anchor byte for one pattern.
+ * @param pattern Pattern name, bytes, and exact-byte mask.
+ * @return A valid anchor when the pattern has a name, bytes, and at least one exact byte.
+ */
+[[nodiscard]] Anchor anchor_of(const Pattern& pattern) noexcept {
     if (pattern.name.empty() || pattern.bytes.empty()) {
-        return false;
+        return {};
     }
-    for (const PatternByte byte : pattern.bytes) {
-        if (byte.exact) {
-            return true;
+    for (std::size_t index = 0; index < pattern.bytes.size(); ++index) {
+        if (pattern.bytes[index].exact) {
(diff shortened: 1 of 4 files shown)


stanuwu:
Thanks. Merged

PR #25, opened 2026-08-15
Add a sword skate module

Haze-xyz:
# Add a sword skate module

A toggle that restores pre-nerf sword skating. **Off by default.**

## What the nerf actually is

One bit of a movement-state field on the player's physics component. It is set while a sword's air
attack is carrying the player, and the glide reads it before anything else and refuses to start.
The jump press is ignored, the speed decays, and the throw cannot be chained.

Clearing that bit on the tick the jump key goes down is the whole change. The client then runs its
own glide.

## How the bit was found

Nothing was disassembled — the image is packed and the movement path is not reachable from any
signature this project anchors, so it was found in data.

I drove the supported build and **79433** (`79433.19.11.13.1925.d2_rc`, 13 Nov 2019, which predates
the 10 March 2020 sword changes) through the same injected inputs at the same timings, sampling the
player's rigid-body velocity every physics tick. Two things came out of that:

- the throw reaches the same speed on both builds
- a glide the client accepts on its own decays at the same rate on both

So neither the throw nor the glide was changed. Only the refusal.

To locate it, I captured memory around the player's physics component at the exact tick the jump is
pressed — once where the glide is accepted, once where a sword throw has it refused — in the same
process at the same addresses. Repeating that and keeping only words that differed every time took
~880 differing words per pair down to 63 candidates. Halving that set with one live test per step
ended on a single bit.

## What the code does

On the physics tick where the configured jump key goes down, if the local player owns the component
and the bit is set, it is cleared. No velocity is written and the player is never moved.

- Only on the press, never on the hold.
- Only that one bit; the attack's other bits are left alone.
- No write when the bit is already clear.
- Unset or wrong jump key, or an open interface: does nothing.

The jump key is configurable because replicated bindings are not what the client reads — it seeds
its own after sign-in. Defaults to Space. Reading the client's key config would remove the field;
happy to add that if you prefer.

## Why a client patch

The refusal is local client state, written and read by the client and never replicated, so there is
no request or push that carries it. I first tried writing the player's velocity to reproduce the
momentum: it gives the speed but not the glide, so no animation, no air control, no landing
behaviour. Clearing the flag is both smaller and correct.

The offset is tied to the supported build, like the other physics-component offsets already here.

## Not in this PR

- The sword rework's other changes. The heavy attack cancels earlier on the current build than on
  79433 — separate mechanic, and a gain rather than a loss.
- Stats or loadout. Jump height comes from mobility, which is the player's own gear.

## Testing

Off: nothing happens, no write. On with the jump key set: works, including with other physics
entities nearby. Wrong key: nothing. Interface open: nothing. Settings survive a relaunch. Teleport
still works alongside, which matters since this shares its hook.

`clang-format` and `clang-tidy` both run clean on the new files. `readability-duplicate-include`
fires, but it fires on existing files too — six times in `teleport_move.cpp`'s include graph — so I
read it as an LLVM version artefact and kept includes explicit.

## Files

New: `client/sword_skate/sword_skate_settings_store.{h,cpp}`,
`client/hooks/sword_skate/sword_skate.{h,cpp}`, `client/ui/sword_skate/sword_skate_panel.{h,cpp}`

Modified: `hooks/teleport/{teleport_lifecycle.cpp,runtime.h,teleport_move.cpp}` (calls the module
from the physics sync, exposes the existing player-ownership test),
`runtime/client_runtime_lifecycle.cpp`, `ui/runtime/client_ui_module_runtime.cpp`, `Sunrise.vcxproj`

## One thing to flag

This reuses the physics sync detour the teleport module installs rather than adding a second one —
the flag is written and read inside that tick, so a frame poll would race. That leaves a movement
feature under the teleport module's hook. If you'd rather that detour became its own player-physics
module first, say so and I'll do it as a separate PR.


stanuwu:
Merged, thanks.

PR #26, opened 2026-08-16
feat: allow custom package loading

zeex64:
Adds package trust handling required to load custom and rebuilt packages:
Bypasses RSA header signature validation.
Accepts rebuilt extended-header and cached-data hashes.
Installs the trust hook before package registration.
Invalidates cached manifest identities when package files change.

_Preserves normal package parsing, bounds checks, decompression, and structural validation._

**Testing**
Verified with custom full packages and patch generations, including rebuilt UI packages and modified package assets.


stanuwu:
Merged, thanks.

PR #18, opened 2026-08-14
PR for Functionality of: Equipment/Item Swap | Currencies/Materials | Shaders | Projections | Finishers | Ships | Sparrows

HiddenInvariant:
## Summary

This pull request adds end-to-end authored inventory and item-action support to Sunrise, including:

- Equipment and item swapping across weapons, armor, Ghosts, ships, Sparrows, emblems, finishers, and other supported equipment buckets.
- Collections acquisition for weapons, armor, cosmetics, and profile-owned items.
- Item dismantling, material returns, and generic equip/unequip flows.
- Shader, ornament, weapon mod, armor mod, Ghost projection, and other compatible socket-plug application paths.
- Dynamic Collections material requirements sourced from installed investment data rather than per-item hard-coded prices.
- Configurable dismantle rewards and atomic profile-material charging/refunds.
- Live QueueZ updates for inventory state, equipped appearance, roster state, and other authenticated peers.
- A version-4 first-load inventory and currency preset for exercising the supported systems.
- A standalone regression-test target for protocol, extraction, appearance, and replication behavior.

## Why

The existing server could describe a configured equipped loadout, but client item actions did not have a complete authoritative mutation and replication path. Requests could finish locally without publishing a durable inventory, socket, material, or appearance change, and some changes remained stale outside the item-details screen.

This change adds the missing installed-content metadata, authored state model, atomic transactions, strict protocol routes, and versioned QueueZ publication needed to make those actions authoritative and observable without reopening menus.

### Runtime-derived investment data

- Extracts collectible-to-item links, item details, socket layouts, compatible plug pools, material requirements, native bucket/equipment-slot mappings, and class-qualified art arrangements from the installed packages.
- Persists the extracted domains in the validated version-33 build-data cache with canonical ordering, cross-domain link checks, bounded capacities, and indexed runtime lookups.
- Fails closed when installed metadata is incomplete or inconsistent instead of publishing a partially usable catalog.
- Keeps runtime behavior content-driven; no per-item compatibility tables or Collections prices are embedded in the server.

### Authored inventory and transactions

- Models account profile stacks plus equipped and unequipped character items with stable instance identities, mutation serials, native bucket placement, quantities, socket state, and item flags.
- Includes every referenced resident object in the initial Family 4 snapshot and rejects dangling SOIDs, duplicate ownership, invalid placement, and capacity overflow.
- Implements prepare/preview/commit transactions for Collections acquisition, profile acquisition, dismantling, equip, unequip, socket changes, and item-state toggles.
[...]

PR #27, opened 2026-08-16
Add subclass equipping and full ability-node selection

Nyxaraa:
## Summary

Three commits, building on each other:

- **Merge PR #18 content, add Linux/clang-cl build support, and enable subclass ability socket-plug handling** — brings in the PR #18 feature set this fork was missing (socket plugs, collectibles, material requirements, opcode 903/1901), adds a CMake/clang-cl/xwin toolchain to build and run under Wine on Linux, fixes two Wine-specific runtime bugs (resource-path backslashes llvm-rc can't resolve, and Winsock hook setup aborting on two GetProcAddress lookups that fail under Wine), and adds an early subclass-aware branch to the opcode-1901 socket-plug path. That last piece turned out to be architecturally incompatible with subclasses (their socket entries aren't real items, so plug-source resolution never succeeds) and is superseded by the dedicated pipeline in the third commit below — left in place here since it's already-pushed history, but opcode 801 is the real path now.
- **Make subclasses equippable through the standard inventory swap** — each character receives its class's other subclasses into unequipped inventory, routed through the existing socket-plug pipeline so they swap via the normal gear-swap UI. Ability picks reset to universal defaults on swap, since the prior picks index into the old subclass's socket-entry list.
- **Add subclass ability node selection (opcode 801)** — the actual missing piece: individual ability-node clicks within an equipped subclass never sent any client request at all before this. Adds an `acquired` socket-entry state (between `ready` and `active`) that turned out to be required before the client will let a node be clicked, a dedicated opcode-801 request/transaction pipeline (staging → commit → item-instance/appearance/roster notifications), and a new `socket_entry_buckets` content domain that resolves each socket entry's real ability-bucket destination from its selector chain during extraction — needed because a bundled multi-node pick (an Attunement) can route its members to different ability slots (e.g. one node changes the super, another the melee) in a way that isn't discoverable from table position or grouping alone.

## Test plan

- [x] Verified in-game: grenade/movement/melee/class/super and all Attunement variants select correctly, survive switching back and forth between options (including options that touch different ability slots), and the picked ability is actually usable in gameplay, not just reflected in the menu.
- [x] Verified subclass swap in and out of inventory resets ability picks correctly and resolves synchronously in both directions.
- [x] Built and ran under Wine on Linux via the added clang-cl/xwin toolchain.


Nyxaraa:
When PR 17 and PR 18 get merged upstream I will clean this PR up, because I don't want to duplicate the changes there.

This unfortunately has to handle PR 17 because I use Linux and PR 18 has some building blocks required for subclass functionality.

zeex64:
lgtm 🔥🔥🔥

Nyxaraa:
New commit added, this is to fix a UI issue, when we tried to have ability choices persists across reloads the game would launch with the server recognizing the default choices but the UI would show the persisted choices, which would then be wrong and redraw when you changed any other ability.

Resetting every ability to default on every login fixes this behaviour, but it means that for now choices can't persists across reboots. That's something I will work on a later time.

Nyxaraa:
Phoenix Dive is broken, in progress of fixing this now.

stanuwu:
Please set this to draft, as its still WIP

Nyxaraa:
Now that upstream includes a lot of the dependencies needed to make this work I've rewritten to include just the subclass changes and all the bugfixes needed, this should be ready to plug and play now. Apologies for the force push over master :)

Nyxaraa:
Resolved ability node picks bleeding over to other subclasses, resolved the conflict after upstream broke state_account_runtime down, merged the bcrypt fix into my fork to reconcile this with latest upstream.

stanuwu:
Merged, thanks.

PR #29, opened 2026-08-16
Fix Wine/Linux cross-compile and runtime compatibility

Nyxaraa:
## Summary
Three small fixes needed to build and run this project under Wine on Linux (via clang-cl/xwin cross-compilation):

- **`sunrise.rc`**: used backslash resource paths (`..\\vendor\\imgui\\LICENSE.txt`), which `llvm-rc` cannot resolve when cross-compiling on Linux. Switched to forward slashes, which both `llvm-rc` and native Windows `rc.exe` accept.
- **`egress_guard_exports.cpp`**: aborted hooking the *entire* Winsock export surface the moment it hit a function `GetProcAddress` couldn't resolve under Wine (`WSAConnectByList`, `GetAddrInfoExA` aren't implemented). Now falls back to small dummy stubs for just those two exports instead of taking down the whole hook set.
- **`dllmain.cpp`**: adds a `GetDC`/`ReleaseDC` priming call before Steam runtime init, needed for Wine's display handling to initialize correctly.

## Test plan
- Cross-compiled with clang-cl + xwin targeting `x86_64-pc-windows-msvc` on Linux
- Deployed and launched under Wine/Proton — confirmed the game boots and connects correctly with these fixes; without them the build either fails to compile (`.rc`) or the hook/init sequence breaks under Wine


TotalTaxAmount:
Oh whoops I didnt see this, I opened #30 with checks to only run the changes under wine and maybe a slightly better placement of the fixes but this would also work

Nyxaraa:
#30 has some good qol changes so I'll go ahead and close this pr in favour of that one. Nice work @TotalTaxAmount

PR #30, opened 2026-08-16
Linux support

TotalTaxAmount:
This PR adds support for running under linux. There are two main things that need to be done in order to allow the game to run under linux.

1. Wine starts in a headless mode with no graphics capabilities and waits until the process tries to create a window or something. Normally this is fine as the game has to create a window. Though in this case it seems DXVK (DirectX to Vulkan) trys to initialize before the game creates its window, causing it to crash if wine hasn't loaded stuff to allow DXVK to query the display which only happens after the process tries to create a window or something.

2. A couple of the functions Sunrise tries to hook via detours are not exported by wine. Namely  `WSAConnectByList` and `GetAddrInfoExA`. I have modified `resolve_specs` to instead hook into some dummy functions instead of these to keep the program stable. (There might be a better way to do this but this seems to work well)

These extra steps/changes are only applied when Sunrise detects it is running under wine/proton. This also contains a few QoL changes for linux devs (fix clangd lsp and have debug builds use debugging symbols gdb can understand)


ocucor:
Tested on Void linux, LGTM 🔥

zeex64:
Tested on Cachyos, LGTM 🔥

Nyxaraa:
Tested on Arch Linux, LGTM 🔥

stanuwu:
Merged, thanks.

ProjectUgarte:
**Sunrise 0.3.2 working on Steam Deck OLED / SteamOS 3.8.16** — tested today, launched as a non-Steam game with Proton Experimental forced, in both Desktop Mode and Game Mode.

- Install done fully on-device: DepotDownloader 3.4.0 linux-x64 (runs natively on SteamOS) with QR login, both depots pulled with `-os windows -osarch 64`, then the 0.3.2 release DLL over `bin/x64/steam_api64.dll`.
- First launch: extraction completed in one go, no relaunch needed. Title screen shows Sunrise 0.3.2.0; loaded a Hunter into the Tower and the EDZ — exploration and weapon firing both work.
- Performance (Game Mode, native 800p): 90 fps in menus (panel cap), steady 60 fps in the Tower/EDZ, GPU ~90% / CPU ~70% uncapped.
- The Wine auto-detection patches from this PR appear to work exactly as designed — zero manual tweaks needed.
- One note for headless/SSH installs: SteamOS sets `KillUserProcesses=yes` via a logind drop-in, so tmux/nohup die on disconnect — background DepotDownloader with `systemd-run --user` instead.

Happy to contribute a full "Installing on SteamOS / Steam Deck" wiki page — I have every step written up and verified on hardware.

PR #33, opened 2026-08-16
misc: update README.md with linux build instructions

TotalTaxAmount:
Update the README with how to build on linux


MaxfieldHutton:
I was thinking I had to dual boot to windows to be able to try this out, I didnt think linux would be possible but it makes sense because we arent connecting to bungie servers so no anticheat to stop us

stanuwu:
Merged, thanks.

PR #35, opened 2026-08-16
Teleport distance increase

Breshi123:
- Increased teleport max distance to 200 units. (up from 100)


Froggy618157725:
Woo, now I don't have to do it XD I've done the same thing locally. For added context, box edge is 100m, so you regularly have an effective death barrier that's 100m thick going downwards.

stanuwu:
Not planned, 100 is a reasonable amount for a normal user. If you want more you can ctrl click into the field to type any amount.

Froggy618157725:
Didn't know about the ctrl click thing. It'd be nice if it was a bit more discoverable, but I don't really have any suggestion for that.

I'd at least push it to 150. Escaping the box barrier would be a lot more convenient that way since it's always 100m, you always have to hit it at least twice. Plus the regularly occurring 100m death barriers.

PR #37, opened 2026-08-16
Replicate FOV, VSync, and keybinding source from settings.json

KyleThmpsn:
## Summary

- Add optional `field_of_view` and `vertical_sync_interval` account settings.
- Add `key_binding_source` with `account` and `computer` options.
- Replicate these settings through the Family-4 account preferences record.
- Close the account seed gate so the client retains the modeled replicated preferences.

These settings are handled together because the target client stores them behind the same account seed gate.

## Why

Sunrise already builds account preferences from `settings.json` and republishes them through account State. However, `accountSeedVersion` remained at version 0, allowing the client’s initial seed pass to take precedence over the modeled preferences behind that gate.

In testing, this caused FOV to return to 85 and Account keybindings to be replaced during sign-in or reload. Sunrise does not currently receive and persist those client-seeded values back into account State, so republishing the account with the gate still open can repeat the reset.

This change makes `settings.json` authoritative for FOV and VSync. Keybinding authority is configurable:

- `account` uses the replicated keybindings from `settings.json`.
- `computer` uses the client’s computer-local keybindings from `cvars.xml` and remains the default.

Graphics settings not modeled in `settings.json` continue to be controlled by the client through `cvars.xml`.

## Settings

- `field_of_view`: 55–105; default `85`.
- `vertical_sync_interval`: 0–4; default `0`.
  - `0` disables VSync.
  - `1`–`4` wait that many refresh intervals.
  - On a 120 Hz display, these correspond to 120, 60, 40, and 30 FPS.
- `key_binding_source`: `account` (`settings.json`) or `computer` (`cvars.xml`); default `computer`.

These fields are optional and use backward-compatible defaults, so the settings schema version remains unchanged.

## Testing

- Built Release successfully with toolset v145 and passed clang-format and clang-tidy.
- Verified the bundled JSON and in-client behavior against upstream `0.2.1-67-g233c811` (commit `233c811`).

### DIFF
--- a/Sunrise/resources/default_settings.json
+++ b/Sunrise/resources/default_settings.json
@@ -228,6 +228,8 @@
           "brightness": 3,
           "show_fps": false,
           "hdr_mode": 0,
+          "vertical_sync_interval": 0,
+          "field_of_view": 85,
           "calibration_primary": 10000.0,
           "calibration_alpha": 0.0
         },
@@ -260,6 +262,7 @@
           "clan_chat_join_mode": 1,
           "chat_auto_hide_mode": 1
         },
+        "key_binding_source": "computer",
         "key_bindings": {
           "fire": { "primary": "left mouse button", "secondary": null },
           "toggle_zoom": { "primary": null, "secondary": null },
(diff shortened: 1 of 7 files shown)


stanuwu:
Merged, thanks.

PR #39, opened 2026-08-17
Orbit slice set changing

Breshi123:
This feature adds the ability to set a default orbit bubble when first starting up (in settings.json) as well as orbit bubble swapping based on last destination visited.

Users will need to provide orbit bubble strings themselves to enter into the settings.json for default orbit. And will also need to provide their own bubble_map.txt file structured with package names and their respective orbit bubbles. This .txt file would go into the Sunrise folder and is setup in a way users can customize what orbit bubble is loaded after each visited destination.

For ex:
tangled_shore = orbit_reef_d2
advent_summer_event = orbit_earth_d2

But you can also get creative with it and do:
tangled_shore = orbit_venus_d2

An orbit_stems.txt file will be created when launching Sunrise with the purpose of referencing which activities in the game belong to which package. (I'm sure there's an easier alternative but this is what I got for now)


stanuwu:
Still considering the best way to go about making this game accurate. Sorry for the delay.

stanuwu:
This is not a correct implementation

PR #48, opened 2026-08-17
Add emote wheel support via the real Emotes collection item

Nyxaraa:
## Summary

Implements the emote wheel by equipping the real, non-equippable "Emotes"
collection item (hash `3183180185`) in place of the individually worn emote,
rather than treating the wheel as something built on top of the worn emote
itself.

- The client already ships a dedicated "Configure" screen for this item and
  opens it on its own once the item is present with valid socket data — no
  new client-side opcode or UI work needed, only correct server data.
- The item's 4 real ordinary sockets hold the 4 wheel slots directly, and the
  client's existing generic socket-plug request (opcode 1901, the same one
  used for weapon mods and shaders) is what the wheel already sends to
  reassign them.
- Ships equipped with 4 default plug hashes from the item's real reusable
  plug pool, so the wheel isn't empty the first time it opens.

## Root cause notes

Unlike every other character-scoped item, this item's real content carries
no native equipment-slot mapping at all. That gap surfaced as four separate,
independent copies of the same "item must have an equipment slot" assumption
across the loadout resolver, the light/power calculation, and the socket
appearance-refresh push — each only failing once the item was actually
equipped (not merely owned), which made the actual cause easy to
misdiagnose as "wrong architecture" before it was isolated. All four now
fall back to the emote slot's own native number for this item specifically.

Also fixes the two ordinary-socket auxiliary hash fields, previously left at
a blanket 0 fill for every item's sockets. They now carry each socket's own
resolved plug definition hash, which is what let the wheel's icon and its
individually socketed emotes actually render instead of silently blanking.

## Test plan

- [x] Emote wheel item renders correctly in the equipped-emote slot
- [x] Opening the item's Configure screen works via the client's own UI
- [x] Reassigning a wheel slot to a different emote via the socket-plug
      request completes without disconnecting
- [x] Existing equipped items (weapons, armor, ghost, ship, sparrow, emblem,
      finisher) are unaffected


Confetti3:
Review: 

The main blocker is the slot-14 fallback. Right now the new logic treats **any** character-scoped item with no native `equipmentSlot` as the Emotes collection item and assigns it slot 14. That behavior appears in the loadout resolver, light resolver, and QueueZ appearance path. This should be explicitly gated to definition hash `3183180185`; otherwise another slotless character item can silently alias the emote slot or cause equipment resolution failures.

The second blocker is boot ordering. `ensure_character_emote_collection()` currently runs during the investment refresh, but Family-4 can be serialized before that refresh executes. On a cache-hit boot, the first Family-4 companion/snapshot can therefore still contain the old individual emote. The migration then changes server-side State later without publishing a corresponding QueueZ mutation/resync, so the client can remain on the stale version for that session. I think the emote canonicalization needs to happen at the same boundary as `ensure_profile_item_identities()` immediately before the first Family-4 account snapshot is taken, and ideally during cache-hit State initialization as well.

There is also a failure-handling issue in `ensure_character_emote_collection()`. If the candidate account fails `account::valid(candidate)` while `failed == false`, the candidate is discarded but the function returns `true`. That can cause the investment worker to consider the migration complete even though nothing was published. The mutation serial increment should also be bounds-checked before casting/incrementing.

A few additional checks would make this safer:

* Verify hash `3183180185` resolves to the expected character-scoped item before migration.
* Verify it has exactly 4 ordinary socket lanes.
* Resolve all four default plug hashes and verify each is allowed in its intended lane.
* If the collection item already exists, validate/repair its socket state instead of skipping it based only on the definition hash.
* Put the “slotless Emotes item => native slot 14” behavior behind one shared helper instead of maintaining the same special case in three places.

I would also add tests for a cache-hit boot, first-run extraction boot, an unrelated slotless character item, an already-present but malformed Emotes item, and migration failure/serial exhaustion.

So my vote would be **request changes for now**. The approach of using the existing generic socket transaction path is good, but the fallback needs to be scoped specifically to the Emotes collection item and the migration needs to be guaranteed before the first Family-4 snapshot is published.

Nyxaraa:
@Confetti3  New commits to address your review :+1: 

- Gate the "no native equipment slot" fallback to the Emotes collection
  item's exact definition hash (3183180185) via one shared helper,
  state::account::inventory::resolve_native_equipment_slot(), used by all
  three call sites instead of three copies of the same loose check. Any
  other slotless character item is now rejected rather than aliasing the
  emote slot.
- Call ensure_character_emote_collection() from
  family4_snapshot_preparer.cpp, at the same boundary as
  ensure_profile_item_identities(), so the migration is guaranteed to run
  before the first (and every full-resync) Family-4 snapshot regardless of
  cache-hit vs. first-run extraction boot ordering.
- Fix ensure_character_emote_collection() returning true when the
  post-mutation account actually failed validation, and add a bounds
  check on the per-character inventory serial before incrementing it.
- Validate the collection item's own definition/detail before migrating
  (no native slot, exactly 4 ordinary socket lanes), validate every
  default plug hash is installed and allowed in its lane, and for an
  already-equipped item validate/repair its real socket contents instead
  of trusting a definition-hash match alone.

Additionally, pushed another commit as the previous defaults were class/race-specific items, not appropriate as
account-wide defaults for an item every character equips. Replaced with
four universal Common emotes. Yes (3184938442), Nope
(48790291), Casual Sit (383973261), and Cheer (2834933816).

Lane order was confirmed empirically in-game (top/bottom/left/right do not
map to lane index 0-3 in link order): lane 0 = top, lane 1 = bottom,
lane 2 = left, lane 3 = right. Final layout: Cheer top, Casual Sit bottom,
Yes left, Nope right.

Nyxaraa:
As for the tests you suggested;

Suite 1 - native, pure logic (9 checks, no game needed)
Tests resolve_native_equipment_slot() in isolation: normal items keep their own slot; negative slots are rejected; an unrelated slotless item is rejected, not aliased to the emote slot, the real Emotes item still falls back to slot 14; a hypothetical future real slot on that item overrides the fallback; and the serial-exhaustion boundary (>= INT32_MAX) is correct at, below, and above the limit.

Suite 2 -  in-DLL, live game data (9 checks)
Tests the "is this item's socket state healthy?" logic against your real account: a correct item is judged sound; wrong policy, wrong plug count, a garbage plug hash, and an empty lane are each independently caught as unsound; an unrelated item is still rejected even with real game data loaded; the real Emotes item and its real default plug pool both resolve correctly against your actual installed content; and the item is confirmed present before the first Family-4 snapshot is built (the boot-ordering fix).

Confetti3:
I reviewed the new commits, and the previous findings were substantially addressed: the slot fallback is now scoped to the exact Emotes collection hash, the shared resolver removes the duplicated special case, the migration validates its build-data assumptions, malformed socket state is repaired, and the serial/failure handling is improved.

I still found three remaining issues.

### 1. [P1] The migration still runs after the first Family-3 snapshot

The migration now runs inside the Family-4 snapshot builder, but the initial subscription sequence builds the Family-3 roster before it builds the unsolicited Family-4 companion.

That means the sequence can still be:

1. Family 3 snapshots the old individually equipped emote.
2. Family 4 runs `ensure_character_emote_collection()` and replaces it with the collection item.
3. Family 0 snapshots the migrated account.

The Family-3 character record can therefore reference the old emote instance while the Family-4 manifest contains the newly generated collection-item instance. No Family-3 correction is published afterward.

The direct Family-0 subscription path also takes an account snapshot without first guaranteeing the migration.

I think the canonicalization needs to move ahead of the entire family-specific snapshot dispatch, rather than living only in `family4_snapshot_preparer.cpp`.

The invariant should be:

> No Family-0, Family-3, or Family-4 image may take an account snapshot until the Emotes collection migration has completed.

A shared preflight at the start of the QueueZ snapshot/subscription path would avoid the three families observing different account versions.

### 2. [P2] Missing or malformed emote build data is still reported as successful migration

`ensure_character_emote_collection()` currently returns `true` when the collection definition/detail or default plug validation fails:

```cpp
if (!resolve_emote_collection_definition(collectionDefinition, collectionDetail)
    || !default_plugs_valid(collectionDefinition.definitionIndex)) {
    return true;
}
```

That causes both callers to treat the invariant as satisfied:

* The investment worker persists the cache and considers extraction complete.
* The snapshot builder continues serializing the account without the collection item.

An omitted detail row, incomplete plug relation, changed layout, or bad default hash therefore silently disables the migration while reporting success.

This should distinguish at least:

* `ready`: migration completed or was already canonical.
* `not_ready`: required extracted data is not available yet.
* `unsupported` or `failed`: the installed content does not match the expected layout.

A result enum would be cleaner because the refresh path and snapshot path may need different policies. At minimum, the snapshot path should not interpret failed prerequisite validation as “safe to serialize the migrated account,” and the failure should be logged with a specific reason.

### 3. [P2] Repairing an existing item clears unrelated item state

The repair path preserves the existing instance SOID, but then reconstructs the entire item from a zero-initialized object:

```cpp
authored_inventory::Item granted{};
granted.instanceSoid = instanceSoid;
granted.definitionHash = authored_inventory::kEmoteCollectionDefinitionHash;
granted.level = 0;
granted.quantity = 1;
granted.mutationSerial = ...;
```

This clears `flags` and resets any other existing scalar state even though the repair only needs to replace malformed sockets.

For an existing collection item, I would copy the current item and mutate only the fields owned by this repair:

```cpp
authored_inventory::Item repaired = present
    ? *collectionSlot
    : authored_inventory::Item{};
```

Then update only the mutation serial and socket policy/count/plugs. The fresh-grant path can continue constructing a canonical new item.

### Verification

The live test coverage described in the comments is useful, but the GitHub Build workflow for the current head still shows `action_required` rather than a passing result, so there is not yet successful CI evidence for the submitted head.

My vote remains **request changes**. The first issue is the remaining blocker: the migration needs to happen before Family 3 and Family 0 can snapshot the account, not between the Family-3 and Family-4 publications.

Nyxaraa:
1 (P1) 
Migration moved ahead of the entire family dispatch. You were right, and I verified the ordering myself: append_queuez_notification builds family 3 via prepare_roster, family 0 reads the account directly at line ~177, and only family 4 ran the migration. New push::ensure_account_canonical() in queuez_account_preflight.cpp now runs before any builder, called from all four public entry points (append_queuez_notification, append_account_resync_notification, append_banner_notification, append_banner_move_notification) plus both resync paths and I removed it from family4_snapshot_preparer.cpp. I cross-checked every account_snapshot() call in the push tree against preflight coverage. The two resync functions were already covered transitively via their familyNActive guards, but I added the call anyway so the invariant holds by construction rather than by reachability argument. It's an atomic load once settled, so the per-frame cost is nil.

2 (P2) 
Outcome enum. ensure_character_emote_collection() now returns EmoteCollectionOutcome  ready / notReady / unsupported / failed. notReady is now a real distinction (build-data domains unpublished, or account not yet valid) separate from unsupported (domains ready but content doesn't match  final for the process, so the preflight latches and stops retrying). Each non-ready outcome logs ev=queuez stage=account_preflight result=skip reason=…. The refresh path only hard-fails on failed, deliberately: letting notReady short-circuit the && chain would block build_data::persist() and break caching for an unrelated reason.

3 (P2) 
Repair preserves item state. Now copies the existing item and mutates only what the repair owns (definitionHash, sockets, mutationSerial); flags and everything else survive. Sockets are still replaced wholesale, since lanes past the used prefix must be empty for the block to validate. The fresh-grant path still builds a canonical item. Worth noting the account is validated whole on entry, so a present item's scalars are already known good  no normalizing needed.

stanuwu:
Merged, thanks

PR #50, opened 2026-08-17
Add a developer console for the settings the pages already expose

Haze-xyz:
Sunrise has three pages that each expose their module's settings, and nothing that lets a line be
typed. This adds a console: one registry of typed commands and variables, an overlay on its own
key, and the eighteen entries the existing modules already had to offer.

## What it is

A module publishes a **descriptor** — a name, a type, its bounds, its help, and the callbacks that
read and write it. Nothing else is declared, because nothing else is needed:

```c
descriptor.name    = "movement.fly_speed";
descriptor.help    = "Flight speed in world units per second.";
descriptor.type    = Type::real;
descriptor.minimum = kMinimumFlySpeed;   // already declared by the module
descriptor.maximum = kMaximumFlySpeed;
descriptor.read    = &read_member<&Settings::flySpeed, Type::real>;
descriptor.write   = &write_member<&Settings::flySpeed, Type::real>;
```

That one declaration then serves four readers: the prompt completes the name, the help prints the
type and range, the parser refuses a value outside it, and the list shows what the module currently
holds.

**A variable carries no storage of its own.** It reads and writes through the module's existing
`get` and `publish`, so the console and the page are looking at one value rather than two copies
that can drift. The bounds are the constants the module already declares and justifies, so a
refusal quotes the same limit the page enforces.

## Entries

| Prefix | Count | What |
|---|---|---|
| `console.` | 3 | `clear`, `find`, `help` — the console answers for itself |
| `log.` | 5 | one threshold per channel, changeable while the process runs |
| `movement.` | 9 | every field the Movement page draws |
| `player.` | 1 | `infinite_ammo` |

`log.*` needed a small addition to the logging module: thresholds could only be set at
initialization. They were already atomic because every writer reads them, so changing one needs no
lock and cannot interrupt a write in progress. The events that explain a problem usually sit below
the level configured before it appeared, which is the whole reason to reach it without a restart.

## Nothing runs on the thread that submits it

A handler reaches state its module owns, and this process already hooks rendering and the network.
Calling one on whichever thread happened to submit it would be a data race waiting for a second
caller, so submitting is separated from running: a checked invocation is queued and runs later on
the thread that drains. A handler runs with no queue lock held, so it is free to take whatever lock
its module needs without ordering itself against this one.

## One thing in here has no consumer yet

Flagging it rather than letting it be found: **`queue::submit` returns a ticket, and nothing reads
it.** The console prints the answer it gets back and ignores which request it belonged to, so today
the ticket only ever serves as the `kNoTicket` a full queue reports.

It is there because a reader is not the only caller this shape is meant for. Anything driving the
console from outside the process has several requests outstanding at once and has to tell the
answers apart, which is the one thing a shared handler cannot do for it. Retrofitting that later
means changing the signature every module's handler is reached through; carrying eight bytes now
does not.

If you would rather it came with its consumer, say so and I will take it out — the queue works the
same without it, and the change is small in either direction.

## Why a handler's answer is not prose

A result carries a sentence for a reader **and** the same answer as named values. The console prints
the sentence and then the rows; anything that is not a reader takes the rows and ignores the prose.
One handler serves both, so the two cannot report different things.

## What it touches outside `core/console`

Ten files, 109 lines, and **no new patch or detour**:

- `core/ui/runtime` — the visibility state gains a second binding, and `interface_open` decides what
  counts as open.
- Three files under `client/hooks/graphics` — they already took one boolean for the cursor policy
  and for polled input. They still take one boolean. Core decides what it means, so neither of them
  learns that a console exists.
- `core/settings` — one more bindable key name, and one more setting parsed. Unknown keys were
  already skipped, so an existing `settings.json` keeps working untouched.
- `core/runtime`, `client/runtime` — publishing and releasing the entries.
- `Sunrise.vcxproj` — the project file names its sources one by one. CMake globs and would have
  built the new ones regardless, which is exactly how the Windows build could have gone out without
  them.

## Deliberately not in this change

- **Making the pages a rendering of this registry.** A descriptor already says everything a widget
  needs: a `bool` is a toggle, a bounded `float` is a slider, a `text` with choices is a combo.
  Doing that would end the duplication between a page and its console entries for good — but it
  would also rewrite pages that work, inside the change that introduces the registry. It belongs in
  a separate proposal, judged on its own, and I would rather offer it than assume it.

## Testing

The parts that do not touch Dear ImGui are exercised off-target: the registry's ordering and
all-or-nothing publishing, the reader's type, bound and choice checks, the queue's ordering and its
behaviour when full, the ring that keeps the newest lines, history, completion, and that a printed
value reads back as the same value. Compile-time checks cover the tokenizer and the folding.

In game: each entry read and written, a refusal outside declared bounds, a threshold changed while
running, and — the one that matters — a value written from the console showing on the page that
owns it.


stanuwu:
Console is not a planned feature. You can attach your own system console during development if needed. Sunrise is not a development tool, and all these features can be toggled through ui with keybinds. A console is useless and adds dead weight and maintainance time.

PR #55, opened 2026-08-18
Inventory fixes: equip ordering, dismantle payouts, Year-1 masterworks

HiddenInvariant:
Bugs reported against 0.3.2.0, four self-contained commits, tested in-game across Red War, Forsaken and Shadowkeep gear.

**Equip swap keeps the displaced item in the clicked cell** (`6164a3b`)
The client orders a bucket's grid by `mutationSerial`, so the displaced item always jumped to cell 0. It now takes the clicked item's prior serial. Same approach as #41 by @joshrad-dev, with the serial's ordering role documented.

**Dismantle pays for energy/heavy weapons and can actually credit** (`6cc137e`)
The payout gate compared the native slot against a semantic-order bound, so only kinetic and armor paid. The shipped profile also seeded payout currencies at their stack caps, so nothing could ever be credited. Starting stacks lowered, Glimmer added, settings layout 7 (files upgrade in place); dropped payouts are logged.

**Rolled socket actions: Year-1 weapon and armor masterworks** (`028dd1c`)
Year-1 masterwork sockets offer only apply/re-roll action plugs; the service answered by socketing a result plug from a roll set that is never in the socket's own pool. Sunrise now does the same, derived entirely at runtime: the plug block is located by its class marker, and its category, roll-set ordinal and linked plug identify actions, weapon results (stat plugs) and armor results (which swap the intrinsic stat perk). Cache format 36.

**Dismantle payout by rarity, gear class and masterwork state** (`5dced03`)
`dismantle_rewards` rows gain optional `rarity`, `class` and `masterworked` filters; unfiltered rows still work. Rarity comes from the item tier byte; "masterworked" is a rolled result or a plug high on its lane's tier ladder (weapon top, armor upper half). Payouts aren't in client data, so the table is settings-driven (layout 8).


stanuwu:
Merged, thanks.

PR #56, opened 2026-08-18
Add toggle to disable AFK timeouts clientside

throgsoft:
Adds a toggle switch to the player UI module that prevents the client from returning a session to obrit after being inactive for too long, and an optional advanced tab that renders the client's session timer, idle timer, and grace period. 

The client keeps one timeout per activity lane. The switch forces every lane to its maximum and the values the client itself authored are captured at runtime so they can be restored when switched off. 

Settings are persisted.

### How
The pointer to the activity timeout values are reached at runtime by resolving the client's getter by the following prologue:

```asm
40 53                  push rbx
48 83 EC 20            sub  rsp, 20h
48 8B 1D 2B 10 1A 02   mov  rbx, [rip+21A102Bh] ; its own global
48 85 DB               test rbx, rbx
0F 84 ? ? ? ?          je   ...
48 89 5C 24 30         mov  [rsp+30h], rbx
E8 ? ? ? ?             call ...
33 C3                  xor  eax, ebx
```

The client then reads its lanes straight off the return value and offset corresponding to the activity lanes from the table below:

| Offset | Activity | Offset | Activity |
| --- | --- | --- | --- |
| `+0xAC` | PvE | `+0xC8` | PvP (matchmade, multiple) |
| `+0xB0` | PvE (guided) | `+0xCC` | PvP (matchmade, single) |
| `+0xB4` | PvE (matchmade, multiple) | `+0xD0` | PvP (special) |
| `+0xB8` | PvE (matchmade, single) | `+0xD4` | PvP (private) |
| `+0xBC` | PvE (special) | `+0xD8` | PvP (Trials) |
| `+0xC0` | PvP | `+0xDC` | Social |
| `+0xC4` | PvP (guided) | `+0xE0` | Orbit |

The lanes are 14 consecutive `uint32` ms from `+0xAC`. No timeout values are hardcoded.

The same object carries a single session 'grace period' at `+0xAC`. No lane can time out until the session has outlived it, hence the minimum timeout offered is a constant 60000 ms.

### How to test
1. Build as usual
2. Navigate to modules -> player -> inactivity
3. Open the advanced tab and enable 'Use set timeouts' (main switch will be forced off)
4. Set social to the minimum value `60000`
5. Load into the tower
6. Alt tab/defocus the game window and confirm you are kicked to orbit in ~60 seconds

### Other Notes
The orbit lane is deliberately guarded and always held at maximum. Setting it too low (less than the 60s grace in my testing) makes the client boot to the title screen resulting in a loop of Marionberry errors, forcing a restart.

### Build confirmation
```PlatformToolset   v145
VCToolsVersion    14.51.36231
Windows SDK       10.0.26100.0
Configuration     Release | x64
Rebuild           full /t:Rebuild, 0 warnings, 0 errors
DLL               machine x64, linker version 14.51
clang-tidy        clean under the repo config
clang-format      -
```

Any feedback is appreciated.


throgsoft:
Demo:
https://github.com/user-attachments/assets/e804f8fc-d10f-44ea-8296-de2bf840d832

stanuwu:
Merged, thanks.

PR #46, opened 2026-08-17
Entity Spawner

ReGlitched:
I fixed it this time dont worry
Adds a way to manually spawn any loaded entity into the world

PR #58, opened 2026-08-18
World population

Adamya08:
This feature adds the automatic spawning of enemies at the destination's own authored positions to make exploration feel closer to the in game experience. It is built on top of PR #46 which is the entity spawner and will not build without it. 
The records do not name the actual combatant, so only the positions of them are used and the enemy type is picked from whatever factions are present on that destination. The placement of the enemies are completely client side.
Unfortunately, enemies cannot be despawned as that is server side which means that all enemies will stay until the player exits the location, also it is not possible to extract on 175 of the 466 locations including the moon as the main one so the world population cannot be filled. I am not sure what is the cause behind this.
Testing
I have so far tested all the EDZ thoroughly and spot checked the rest of the planets. All of them worked apart from the moon where extraction isn't possible


Adamya08:
There is someone in discord who has a server side enemy implementation in the works, I am closing this due to it being obsolete

NotJeppy:
This is literally the pull request I mentioned getting this from. This was not my work. I said so when I sent my video. Please reopen this.

PR #60, opened 2026-08-19
Fix MSVC heap exhaustion in physics state storage

Confetti3:
## Summary

- remove nested default member initializers from the fixed physics state storage records
- initialize the global storage aggregate explicitly with the same initial values
- avoid generating the oversized `Storage` initialization function that causes MSVC error C1060

The storage is cleared with `SecureZeroMemory` during initialization, so this keeps the existing runtime state while allowing the Release build to compile.

## Verification

- `msbuild Sunrise.sln /m /t:Rebuild /p:Configuration=Release /p:Platform=x64 /verbosity:minimal`
- deployed Release x64 build tested successfully against the target game build

Upstream failure: https://github.com/stanuwu/Sunrise/actions/runs/32205625456


stanuwu:
Merged, thanks.

PR #63, opened 2026-08-19
Fix missing MSVC project source entries

Confetti3:
## Summary

Adds the source and header entries missing from `Sunrise.vcxproj` for:

- socket entry bucket catalog
- opcode 801 codec

Without the two compilation units, the `Release|x64` MSVC build fails with three unresolved external symbols.

## Validation

- Clean `Release|x64` MSBuild rebuild succeeds with the repository's `v145` toolset.
- `git diff --check` passes.


Confetti3:
Closing because the missing MSVC source entries are now present on upstream master. PR 57 has been rebased onto that fix.

PR #68, opened 2026-08-20
feat: Add persistence across user sessions

ocucor:
## Persist runtime inventory and equipment across restarts

Every change a player makes at runtime is currently lost on exit. Equip something,
dismantle something, insert a mod, and the next boot rebuilds the account from
`settings.json` as though none of it happened. This adds a small binary overlay beside
`settings.json` holding the mutable half of the account, applied over the authored one at
startup.

`settings.json` stays the baseline and is never written to. Deleting `runtime-state.bin`
restores the authored account exactly.

### How it works

`state::persistence::load()` runs once from `core_runtime` after every boot stage
succeeds. `state::persistence::save()` runs at the end of the seven commit functions that
mutate the account: equipment swap, item and profile acquisition, dismantle, socket plug,
item state, and subclass selection. Saving from the commit rather than the request means a
refused mutation never reaches the file. Each call sits after the state lock is released,
since SRW locks are not recursive and `save()` takes a snapshot.

Persisted per character: equipment slots, the unequipped item array and its count, the
next inventory serial, `acquiredSubclassAbilityMask` and `lastOrbitedDestination`.
Account-wide: the profile item rows and their count. Identity and appearance are left
alone.

### The file

Fixed-size packed records with a magic and version in the header, written field by field
rather than by copying the in-memory structs, so the state model can change shape without
silently invalidating files. Written to a `.tmp` sibling and renamed over the target, so a
crash mid-write costs the newest save and not the file the next boot reads. Characters and
the account are matched by SOID and never by position, so reordering authored characters
cannot hand one character another's loadout.

`Sockets::plugCount` is stored explicitly rather than inferred from which plugs are
present. `valid(Sockets)` rejects any plug at or past the count, and a count may
legitimately cover a socket whose plug is empty, so neither can be derived from the other.
Restoring without it fails validation on every item carrying a mod.

### Failure handling

The overlay never stops a boot. A missing, short, oversized, mismatched or invalid file is
logged and skipped, leaving the authored account in place: `absent` on a first run,
`fail_layout` on a header or size mismatch, `fail_invalid` when the result does not pass
`account::valid`, and `ok` with the version and character count otherwise. The write side
reports `fail_open`, `fail_write` and `fail_replace` separately.

### Testing

Built Release x64 against v145 (and v143 for those still on VS 2022). Verified the round trip by equipping across several
characters, quitting, and confirming the loadout returned on the next boot. Also confirmed
that deleting the file falls back to the authored account and that an older layout is
refused rather than misread. Equip, dismantle and socket paths have been exercised
directly; profile acquisition and subclass selection are wired identically but have had
less time in front of them.

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -376,6 +376,7 @@
     <ClCompile Include="src\steam\interfaces\tables\table_runtime.cpp" />
     <ClCompile Include="src\steam\interfaces\steam_interface_factory.cpp" />
     <ClCompile Include="src\state\runtime\state_runtime.cpp" />
+    <ClCompile Include="src\state\runtime\persistence\runtime_state_file.cpp" />
     <ClCompile Include="src\state\activity\activity_session_lookup.cpp" />
     <ClCompile Include="src\state\activity\activity_world_arrival.cpp" />
     <ClCompile Include="src\state\activity\membership\activity_membership_query.cpp" />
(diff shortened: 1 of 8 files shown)


stanuwu:
Not planned, wait for sqlite migration

PR #69, opened 2026-08-20
Add No Turnback and Godmode player options

ltsReaver:
## Summary

- add a persistent No Turnback toggle to the Player tab
- add a persistent Godmode toggle to the Player tab
- keep each feature in its own hook module
- validate original instructions before patching
- restore original instructions when disabled or during shutdown

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -227,6 +227,8 @@
     <ClCompile Include="src\client\hooks\sword_skate\sword_skate.cpp" />
     <ClCompile Include="src\client\hooks\fly\fly.cpp" />
     <ClCompile Include="src\client\hooks\infinite_ammo\infinite_ammo.cpp" />
+    <ClCompile Include="src\client\hooks\no_turnback\no_turnback.cpp" />
+    <ClCompile Include="src\client\hooks\godmode\godmode.cpp" />
     <ClCompile Include="src\client\input\window_focus.cpp" />
     <ClCompile Include="src\client\hooks\teleport\teleport_lifecycle.cpp" />
     <ClCompile Include="src\client\hooks\teleport\teleport_move.cpp" />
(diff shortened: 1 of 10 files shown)


Qutsn:
I tested both features from this PR in-game.

Tested DLL SHA-256:
`FE7EE2216989216B81DE76295CE0893FE1E9A02B4FD5ABE160E67598CB23E187`

## No Turnback

When tested without `Godmode`, `No Turnback` works as expected.

Enabling it while inside a `Turn Back!` zone immediately removes the warning
and countdown. Disabling it while still inside the zone brings the countdown
back, and the player dies normally when it reaches zero.

## Godmode

`Godmode` does not take effect immediately. The player still dies the first
time, and the feature only begins affecting later deaths after the first
respawn.

### Reproduction steps

1. Enable `Godmode`.
2. Jump off a cliff or die by another ordinary method. The player still dies
   normally the first time.
3. Respawn, then disable `Godmode` without changing activities or restarting
   the game.
4. The player remains stuck with only a sliver of health.
5. Passive health/shield regeneration does not start.
6. Crossing an OOB death barrier no longer kills the player.
7. A `Turn Back!` countdown reaches zero and disappears, but the player remains
   alive.
8. Trigger a healing ability or another active healing effect.

The active healing effect immediately refills the health bar. After that,
passive regeneration works again and normal deaths can be triggered again.

### Live-process verification

After disabling `Godmode`, I checked the instruction in the live process. The
patched bytes had already been restored from:

`83 CE FF 90`

to the original bytes:

`8B 74 24 38`

The saved setting also showed `godmode_enabled: false`.

I manually called `FlushInstructionCache` for the restored instruction as a
diagnostic step, but the issue remained. This rules out the patch simply being
left enabled or the CPU continuing to execute stale instructions.

The instruction bytes are restored correctly, but restoring them does not undo
the health and death behavior created after the first respawn. An active
healing effect is currently required to return the player to normal.

### Expected behavior

- `Godmode` should take effect without requiring the player to die once first.
- It should not leave the player stuck at near-zero health or prevent passive
  health/shield regeneration.
- Disabling it should immediately restore all normal healing and death
  behavior, without requiring an active healing effect, activity reload, or
  game restart.

## PR scope

The contributing guidelines state:

> **One Feature** - Do not put multiple features into one PR.

`No Turnback` and `Godmode` are separate features and should be submitted as
separate PRs. This would also allow the working `No Turnback` implementation
to be reviewed independently from the issues with `Godmode`.

stanuwu:
Please use detours instead of raw patches
Also do not write game memory from UI thread

PR #70, opened 2026-08-20
Feat: Character creation and Deletion

Toucan-t:
This adds functionality to the in game character menu, adding character creation, deletion and customization.
Characters are saved as they normally would be in the settings.json and are updated in real-time.
Customization options applied to the character such as race or hair are persistent.
(Face markings are not yet functional)


Toucan-t:
I'm not happy with this anymore

PR #72, opened 2026-08-20
Extend field of view limit to 155

ltsReaver:
allows setting the field of view in the settings.json to above the 105 cap since the game actually supports over 105 and allowing over 105 in the config allows users to manually set it over the ingame cap if they wish.


stanuwu:
Merged

PR #73, opened 2026-08-20
Add selectable coordinate fly mode

ltsReaver:
added selector between velocity and a teleport styled fly due the fact I want to fly without having the no clip enabled as I find the no clip quite doggy and flying with changing coords is better imo than using the no clip


stanuwu:
Not planned

PR #71, opened 2026-08-20
Implement WS-701 settings state updates

gagefulwood:
## Summary
- added a bitwise decoder for the ws opcode 701 and the reflected schema 0x80807603.
- extracts supported account preferences and keyboard/mouse bindings into `SettingsDelta`
- applies the settings changes via a state prepare/commit boundary with validation.
- carries the prepared settings through web service and the encrypted BAP transactions.
- publishes account/settings mutations after response success.
- consolidated preference indices, wire encodings, and semantic assignments into a schema table.
- moves the native keybinding slot map into the neutral State settings for shared inbound/outbound usage.
---
##### `Sunrise/src/middleware/web_service/messages/opcode701/opcode701_codec.h`
- Defines the public interface for the WS-701 decoder.

##### `Sunrise/src/middleware/web_service/messages/opcode701/opcode701_codec.cpp`
- Reads incoming WS-701 packets and pulls the user's settings and keybindings.

##### `Sunrise/src/state/account/settings/settings_delta.h`
- Defines updates for absent fields to be distinct from false or zero values present in the payload.

##### `Sunrise/src/state/account/settings/settings_delta.cpp`
- Merges the settings the user changes/altered without overwriting all of the data currently recorded.

##### `Sunrise/src/state/runtime/state_account_settings_runtime.cpp`
- Validates new settings, locks the state, then saves changes.

##### `Sunrise/src/state/account/settings/native_key_binding_map.h`
- Shared keybind dictionary mapping to the 60 native keybind slots present to internal actions. Moved it to this neutral location for better inbound and outbound implementation/accommodation.

##### `Sunrise/Sunrise.vcxproj`
- Updated the Visual Studio config file for future builds.
##### `Sunrise/src/core/settings/state/key_bindings_parser.cpp`
- Cleaned things up a bit and added safe handling. 

##### `Sunrise/src/state/account/settings/key_bindings.h`
- Defines a  shared mask and constants for native input indexes, modifiers, and one unbounded sentinel.

##### `Sunrise/src/state/account/settings/settings_state.h`
- Added equality rules for server to check if state changed before saves. 

##### `Sunrise/src/state/account/settings/settings_state.cpp`
- Validates incoming settings.

##### `Sunrise/src/middleware/datagen/family4/account/preferences/preferences_encoder.cpp`
- Updated the outbound data generator to utilize the new shared keybind dictionary.

##### `Sunrise/src/server/web_service/web_service_actions.h`
[...]

PR #75, opened 2026-08-20
replaces profile_setup_skip's hook with the server side equivalent

ehkogh:
This does the base work on the server side to skip over profile_setup's state without the client hook. This should open the door for at least some people adding the rest of the codec for ID 701 as that should be where all the settings exist.

I updated `default_settings.json` to remove the profile_setup hook and enable `profile_setup_completed` on your account state.

Worst case scenario you can do this in your own `settings.json`:

```json
"client": {
  "skip_profile_setup": false,
  ...
}
"state": {
  "account": {
    "profile_setup_completed": false,
    ...
  }
}
```

---

I did see that #71 does add a lot of the 701 codec, but doesn't add profile_setup as well.

### DIFF
--- a/Sunrise/resources/default_settings.json
+++ b/Sunrise/resources/default_settings.json
@@ -37,6 +37,7 @@
     },
     "suppress_peer_relay": true,
     "fade_release": true,
+    "skip_profile_setup": false,
     "region_private": false,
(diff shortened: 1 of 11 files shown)


stanuwu:
Merged, thanks

PR #80, opened 2026-08-21
feat: basic roster functionality

ocucor:
### What

Answers the family-two subscription, which was previously served with an empty
snapshot, so the panels drew a name and a blank emblem.

### Why

The panel row resolves the emblem with two lookups, not one:

1. slot 0, keyed by the account soid, gives a directory
2. slot 1, keyed by the qword at directory `+8`, gives the member record
3. the emblem definition index is read from member `+36`, its variant from `+38`

Both objects have to be resident at the same time. A full snapshot prunes every
object it does not name, so publishing one slot per message can never satisfy
that chain whichever slot is chosen

The variant is always sent as `0xFFFF`. The reader tries the variant first and
falls back to the definition index only on that sentinel, so a real number there
resolves art against a bogus variant entry.

The published index tracks the live loadout rather than a constant. The client
resolves this account-keyed object as *the account's* emblem, so a fixed index
pins it globally. character select, inventory and orbit all stop reflecting an
equip while the equip itself keeps succeeding.

That is also why an emblem equip re-arms a delayed second copy: the snapshot is
built at subscribe time, so a mid-session change would otherwise stay stale until
relaunch. The re-arm reuses the root the subscribe was answered against, is gated
on the emblem slot so unrelated equips don't spend a push, and carries its own
flag rather than reusing another family's arm.

### Effects

- The Roster and Fireteam panels draw the account's real emblem.
- Changing emblem updates the panels live, without a relaunch.
- Server-side only. No client patches, no config or settings changes.

### Testing

Confirmed in game: the emblem renders.

### What's Next?

I wasn't able to fit it here, so another PR will be opened in the future for the following:
- Power scalar updates in-state
- Power level updates in roster, orbit and inventory view
- Current activity in roster.

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -903,6 +903,7 @@
     <ClCompile Include="src\server\bap\encrypted\push\snapshot\initial_snapshot.cpp" />
     <ClCompile Include="src\server\bap\encrypted\push\snapshot\banner_snapshot.cpp" />
     <ClCompile Include="src\server\bap\encrypted\push\snapshot\roster_snapshot.cpp" />
+    <ClCompile Include="src\server\bap\encrypted\push\snapshot\social_roster_snapshot.cpp" />
     <ClCompile Include="src\server\bap\encrypted\push\snapshot\snapshot_storage.cpp" />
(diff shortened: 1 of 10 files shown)


stanuwu:
Merged, thanks

PR #83, opened 2026-08-21
Add optional setup and dev-loop helper scripts

DatSwagBoi:
## What this adds

A `scripts/` directory with three PowerShell helpers and one Python documentation generator. Nothing in the build, the solution, or the CMake configuration references them — they are pure conveniences and the repository behaves identically with the directory deleted.

**`SunriseInstall.ps1`** — first-time setup. Checks for Git, MSBuild/Visual Studio and the Windows SDK, installs Git via winget if it is missing, and *offers* Visual Studio rather than installing it unasked. Adds Defender exclusions for the game and work folders, fetches the game with DepotDownloader, installs the latest release DLL as a rollback baseline, clones the repo and builds it once, then writes a deploy script pointed at that machine's folders. Re-runnable: completed steps are detected and skipped, so a failed run can simply be repeated.

**`Sunrise-DevLoop.ps1`** — build, deploy, launch and attach the debugger in one command. The repo is found from the script's own location and MSBuild through `vswhere`; the game folder is asked for once and remembered. Flags cover `-BuildOnly`, `-NoAttach`, `-Config Debug`, `-OpenIDE`, and `-Rollback official|vanilla` to restore a baseline DLL.

**`gen_glossary.py`** — generates a browsable index of the codebase from the doc-comments the headers already carry, grouped by module. Python standard library only, no dependencies.

**`README.md`** — usage for all of the above.

## Why

The README documents the build and deploy steps clearly, but a first-time contributor still runs them by hand every cycle: rebuild, close the game, copy the DLL and PDB, relaunch, re-attach. These automate exactly those documented steps and nothing more.

## Effects and risk

- **No behavioural change to Sunrise itself.** No source file, project file or build input is touched. The diff is four new files under a new directory.
- **No copyrighted data enters the repository.** DepotDownloader runs against the user's own Steam account and prompts for its own credentials and Steam Guard; the scripts never handle, store or log them. The Steam account *name* is a parameter, nothing more.
- **The one destructive operation is opt-in.** Removing an existing clone happens only behind an explicit `-ForceReclone` flag, after a warning.
- **Elevation is only for Defender exclusions.** Without it the installer warns and continues rather than failing.
- Depot manifests go stale, so they are parameters (`-Depot1Manifest` / `-Depot2Manifest`) rather than values to edit in the file.


stanuwu:
No, not planned
Everyone can use their own scripts as needed i wont ship any

PR #84, opened 2026-08-21
Implement Triumphs, lore collectibles, titles, seasonal progression, Artifact perks, bonus power, and reward publication

Nyxaraa:
Implements claiming Triumphs through Web Service opcode 1801, with claims persisted across restarts and Triumph Score totalled server-side.

Also adds lore book support: presentation node extraction, category progress counting, visibility gates resolved from each category's unlock expression, and node table persistence so a warm start still counts.

Verified in game against build 86657.


Sunrise/src/state/record_claims/record_claims.cpp
```diff
@@ -0,0 +1,394 @@
...
+    if (written > 0) {
+        core::log::write(core::log::Channel::state,
+                         std::string_view{result} == "ok" ? core::log::Level::info
+                                                          : core::log::Level::warn,
+                         {line.data(), static_cast<std::size_t>(written)});
+    }
+}
+
+/** Writes every held claim. The caller holds the lock. */
+void store_locked() noexcept {
```
yazan-albaiz: In its current state, this can't report a write failure.

Sunrise/src/server/bap/encrypted/body/bap_service_body.cpp
```diff
@@ -451,6 +452,23 @@ bool process(const ServiceRoute& route,
...
+            if (!middleware::web_service::encode_response(
+                    message,
+                    middleware::web_service::ResponseShape::statusPair,
+                    status,
+                    output,
+                    written)) {
+                core::log::write(core::log::Channel::server,
+                                 core::log::Level::warn,
+                                 "ev=ws1801 stage=response result=fail");
+            }
```
yazan-albaiz, line 671: Here it can also ignore reply failure.

Sunrise/src/server/bap/encrypted/queuez/queuez_outcome_staging.cpp
```diff
@@ -369,6 +369,16 @@ bool stage_service_outcome(Scratch& scratch,
...
         after = acquisition.after;
+    } else if (outcome.hasRecordClaim) {
+        // A claim rewrites one byte of the account flag bank and leaves the manifest alone, so a
+        // full account snapshot at the next version carries it with no other staging.
+        if (!push::append_account_resync_notification(
+                scratch, before, key, nonce, response, written, after)) {
+            core::log::write(core::log::Channel::server,
+                             core::log::Level::warn,
+                             "ev=ws1801 stage=queuez_resync result=fail");
+            return true;
```
yazan-albaiz, line 711: And here it can ignore push failure.

Sunrise/src/state/build_data/records/definition.h
```diff
@@ -0,0 +1,49 @@
...
+ * as 102115. It is a plain replicated value, not a progression and not derived by the client, so a
+ * server that wants a score has to total one itself.
+ */
+inline constexpr std::uint16_t kTriumphScoreValueIndex = 2115U;
+
+/** A record whose completion flag no mapping table addresses carries this instead of an index. */
+inline constexpr std::uint16_t kUnavailableFlagIndex = 0xFFFFU;
+
+/** A record naming no category value slot carries this instead of an index. */
+inline constexpr std::uint16_t kUnavailableValueIndex = 0xFFFFU;
```
yazan-albaiz, line 27: nit: can these not be extracted at runtime?

Sunrise/src/state/build_data/nodes/definition.h
```diff
@@ -0,0 +1,89 @@
...
+inline constexpr std::uint16_t kLoreNodeFirst = 815U;
+inline constexpr std::uint16_t kLoreNodeLast = 854U;
+
+/** @return True when this node is a lore book category, the only kind this build counts. */
+[[nodiscard]] constexpr bool lore_category(std::uint16_t definitionIndex) noexcept {
+    return definitionIndex >= kLoreNodeFirst && definitionIndex <= kLoreNodeLast;
+}
+
+/** A node whose gate names no addressable flag carries this instead of an index. */
+inline constexpr std::uint16_t kUnavailableFlagIndex = 0xFFFFU;
```
yazan-albaiz, line 28: same "runtime" nit

Sunrise/src/state/build_data/cache/records/cache_investment_records.cpp
```diff
@@ -78,4 +78,27 @@ bool decode(const ProgressionRecord& record, progressions::Definition& value) no
...
+/** Decodes one record after checking its padding. */
+bool decode(const RecordDefinitionRecord& record,
+            build_data::records::Definition& value) noexcept {
+    value = {};
+    if (record.reserved != kReservedFieldValue) {
+        return false;
+    }
+    value = {record.definitionIndex, record.completionFlagIndex, record.scoreValue};
+    return true;
+}
```
yazan-albaiz: This omits `categoryValueIndex` which could publish stale or incomplete data.

Sunrise/src/middleware/datagen/family4/character/character_encoder.cpp
```diff
@@ -157,6 +162,23 @@ bool encode(const state::CharacterState& state,
...
+            && state::build_data::nodes::load_and_publish()) {
+            nodesPublished.store(true, std::memory_order_relaxed);
+        }
+    }
+
+    // One lore book counts in the character bank rather than the account one.
+    (void)state::record_claims::apply_character_node_progress(object.objectiveValues);
+
+    // One lore book's gate is character scoped rather than account scoped.
+    (void)state::build_data::nodes::apply_character_visibility(object.acquiredFlags);
```
yazan-albaiz, line 300: This look like claims progress first, then overwrites every objective value. Worth taking a look.

Sunrise/src/middleware/content/packages/tables/unlock_expression.h
```diff
@@ -0,0 +1,87 @@
...
+        return false;
+    }
+    const std::size_t pointerAt = rowAt + field + 8;
+    const std::int64_t target = static_cast<std::int64_t>(pointerAt) + relative
+                                + static_cast<std::int64_t>(kHeaderSkip);
+    if (target < 0
+        || static_cast<std::size_t>(target) + static_cast<std::size_t>(count) * kUnlockInstructionStride
+               > table.size()) {
+        return false;
+    }
```
yazan-albaiz, line 52: can overflow before its check

Sunrise/src/state/build_data/nodes/node_catalog.cpp
```diff
@@ -0,0 +1,114 @@
...
+    return g_definitions.count();
+}
+
+/** Calls back for every node, under the shared lock. */
+void for_each(void* context, void (*visit)(void*, const Definition&) noexcept) noexcept {
+    const Lock::Shared guard(g_lock);
+    for (const Definition& node : g_definitions.rows()) {
+        visit(context, node);
+    }
+}
```
yazan-albaiz, line 76: seems to call unchecked function pointers

Sunrise/src/state/build_data/cache/records/format.h
```diff
@@ -236,6 +238,15 @@ struct ProgressionRecord {
...
 };
 
+/** Disk form of one record and the account flag bank row its claim sets. */
+struct RecordDefinitionRecord {
+    std::uint16_t definitionIndex{};
+    std::uint16_t completionFlagIndex{};
+    std::uint16_t scoreValue{};
+    /** Must be zero, so the packed record row always matches. */
+    std::uint16_t reserved{};
+};
```
yazan-albaiz, line 289: omits `categoryValueIndex` which could cause warm-cache parent counts to lose record data

yazan-albaiz (commented):
Solid work overall!

My main concern is that zero progression should not be hard-coded as Sunrise’s default. Triumph, score, and lore progress should be configurable through settings.json, the Sunrise menu, or later a server config of sorts, so clients can opt into full progression.

Nyxaraa:
Moved to Draft as I started working on collectables and realized they're somewhat entwined, so I'm now working on further implementation.

SharXeniX:
I have been reading the same tables from a pre-Beyond-Light depot, and a few things turned up that
may be useful here. It looks like the same content you verified against: your code says the shipped
build declares 2242 records and 924 presentation nodes, and that is exactly what my depot counts,
and your lore range 815 to 854 lands right in its native numbering. So the row numbers below should
transfer as they are rather than needing a shift. That said, this is my reading and I may well have
misread something, so treat it as a lead rather than a correction.

**On the parent bars, I may have found where they read.** My reading suggests a record's bar
shows the progress of its **objective**, and that the objective names the value slot carrying it.
What looks like the objective table sits at investment root **slot 58**, in rows of **160 bytes**:

| offset | what I think it is |
|--:|:--|
| `+0` | the objective hash, the one the shipped manifest publishes |
| `+8` | an expression: a count, a self-relative offset, then instructions |
| `+48` | the completion value |

The native row index appears to equal the shipped manifest's objective index; I
checked the hash at `+0` on four rows only (5153 -> `0xECD525DC`, 4698 -> `0x2208ADDD`, 4788 ->
`0xB301F58D`, 5442 -> `0x59F5275E`), which is thin but consistent.

The expression at `+8` was a single instruction everywhere I looked, and one of two shapes:
`VAL(slot)` where the count seems to live, or `FLAG(slot)` where the objective looks boolean. If
that reading is right, mapping the slot through the value bank at root slot 113, the way a flag slot
goes through 111, would give the row to write.

Applied to three lore book parents, it gives:

| parent record | its category's slot | what `+1` would give | what its objective seems to name |
|:--|--:|--:|:--|
| The Pigeon and the Phoenix (1983) | 7060 | 7061 | `VAL(7061)`, so `+1` would be right here |
| Letters from a Renegade (1422) | 8275 | 8276 | `VAL(5128)`, nowhere near |
| Eva's Journey (1490) | 10343 | 10344 | `FLAG(18740)`, not a count at all |

Which would suggest `+1` happens to hold on some books rather than being the rule, and might explain
why marking two of them confirmed it while the contiguous ones drove the wrong bar. I have not
tested any of this against a running client, so it may be wrong in a way I cannot see from the
packages alone.

Two smaller observations, offered the same way:

  Chapter records seem to carry a second expression at **`+136`**, `VAL(own_slot) CONST(1) >=`, which
  looks like it names the chapter's own progress slot directly. Field `+120`, the one read here for
  the category, also parses as a three-step condition (`VAL(slot) CONST(0) cmp`) rather than a plain
  reference; taking its first `VAL` operand appears to give the same slot either way.

  Two cheap checks that need no client, if they are useful: the score at record `+92`
  reproduces the shipped manifest's `completionInfo.ScoreValue` for 2212 of the 2233 records that
  manifest publishes, the misses being past row 2185 where the manifest's own numbering seems to
  drift; and a node's native row appears to be the manifest's presentation node index **plus four**,
  with 136 of the 139 nodes that own records owning exactly what the manifest lists. Both caught
  stride mistakes for me.

One landmark ties the objective reading to a running client: the Jade Rabbit triumph's objective `0xECD525DC` resolves through `VAL(6202)` to account value
row 2503, and writing that row does move its counter in game for me.

---

Unrelated to the records, but next door, and I spent a while on it: **which of the
five mapping banks a flag lives in seems to be decided by scope rather than by what the flag is
about.**

Checklist entries carry their own scope, and it lines up with the bank on all 620 of them:

```
entry scope 0  ->  bank 0        entry scope 1  ->  bank 2
```

Ghost Lore, the Cayde journals, Corrupted Eggs, Cat Statues, Latent Memories, Sleeper Nodes,
Calcified Light, Ahamkara Bones, Dead Ghosts and Savathun's Eyes are scope 0 and fall entirely in
bank 0; Lost Sectors, Adventures and the Jade Rabbits are scope 1 and fall entirely in bank 2. The
Region Chests checklist is the one that made me believe it: its 111 entries are 100 at scope 0 and 11
at scope 1, and they split 100 in bank 0 and 11 in bank 2. Which would mean the same kind of
collectible can be remembered per account or per character.

If that is right, **bank 2 would be the character bank**, written into the character object's own
flag bank of 4096 slots rather than the 256-slot character block of the account object. What I can
say is that writing rows 674 to 679 in the character scope removed the first six Jade Rabbits from
the Moon for me, with the last three left at zero staying visible. Worth knowing either way: its
rows run well past 256, Lost Sectors sitting at 2187 to 2271, so anything assuming the 256-slot
block would write them and have them dropped, which looks from the outside exactly like a flag that
does nothing.

Nyxaraa:
# Summary

  This PR adds functional Triumph claiming, accurate lore progression, equippable Titles, collectible rewards, Lost Ghost handling,
  Seasonal XP, Season Pass claiming, and reliable Queuez acquisition presentation.

  It also substantially restructures and cleans the supporting record, progression, inventory, and BAP publication code.

  ## Triumphs and record progression

  - Added support for Triumph claim requests through opcode 1801.
  - Extracts record definitions, completion flags, score values, objectives, and rewards from installed build data.
  - Persists claimed and claimable Triumph states across restarts.
  - Publishes claimed records back into the account flag bank.
  - Calculates and displays total Triumph score.
  - Updates category counters and parent progress bars immediately.
  - Supports partially completed, completed-but-unclaimed, and claimed objective states.
  - Grants authored Triumph item rewards through the normal inventory acquisition flow.
  - Prevents already-claimed Triumphs from being claimed twice.
  - Preserves client-derived claimability where the server does not own the original completion source.

  Implementation: records and presentation nodes are extracted into build-data catalogs. Claims are staged transactionally, published
  through Family 4, and only committed after the resulting account update succeeds.

  ## Lore books and chapter counting

  - Correctly reveals lore books and individual chapters.
  - Separates chapter collection state from chapter claim state.
  - Maps parent progress bars to their actual objective value slots.
  - Prevents one book’s progress from overwriting an adjacent book.
  - Supports account-scoped and character-scoped lore categories.
  - Handles hidden and single-slot lore category gates.
  - Correctly excludes parent records from normal chapter totals.
  - Adds special handling for the four books that grant chapters directly from their shared counter:
      - Stolen Intelligence
      - Unveiling
      - A Man with No Name
      - Revelation

  - Supports books where collection reveals the chapter but claiming advances the visible bar.
  - Persists lore claimability, claims, partial objective progress, and score.

  Implementation: extracted node relationships and objective mappings are used to derive each book’s children, gate, counter, and parent
  bar rather than relying on broad bubble associations.

  ## Exact lore collectible resolution

  Collectibles now resolve to their specific authored lore entry instead of simply granting the next unowned chapter.

  Supported collectible families include:

  - Dreaming City Awoken Memory Crystals
  - Ahamkara Bones
  - Corrupted Eggs
  - Dead Ghosts and Ghost Stories
  - Fallen Prince drones
  - Menagerie Confessions vases
  - Reckoning/Dust collectibles
  - Moon Lost Ghosts

  Implementation:

  - Uses authored SObject record rows where available.
  - Uses stable collectible ordinals for families with contiguous definition ranges.
  - Uses explicit ordinal mappings where the record range is non-linear.
  - Uses measured world positions only for generic objects that carry no individual identity, such as Menagerie vases and Dust objects.
  - Checks primary and additional incident targets before applying positional fallback.
  - Duplicate collection attempts do not duplicate lore progress.

  ## Corrupted Eggs and Dreaming City cats

  - Corrupted Eggs grant their associated lore/Triumph progress.
  - Eggs award a random Dreaming City item.
  - Cat statues advance Remember Your Manners.
  - Cats use the same Dreaming City reward pool.
  - The reward pool contains:
      - Dreaming City weapons
      - Class-appropriate Reverie Dawn armour

  - Corrected invalid and mismatched hashes that previously caused generic weapons or weapon-heavy results.

  Implementation: collectible incidents enqueue an installed-build item definition through the standard deferred world-reward acquisition
  path.

  ## Jade Rabbits

  - Jade Rabbit interactions advance Let Them Eat Rice Cakes.
  - Grants a random Moon world reward.
  - Supports Moon weapons and class-appropriate Dreambane armour.
  - Added the required pursuit/quest inventory support needed for Rice Cakes and Lost Ghost testing.

  ## Moon Lost Ghosts

  - Moon Ghost objects map to their exact Luna’s Lost lore entries.
  - Collection advances Luna’s Lost Are Found.
  - Completing a Ghost search removes the corresponding pursuit as expected.
  - Grants nine Phantasmal Fragments.
  - Grants 2,500 unmodified Seasonal XP.
  - Publishes the XP gain through the gameplay HUD.
  - Duplicate Ghost collection does not repeat lore progress or rewards.

  ## Equippable Titles and Seals

  - Added Title equip/unequip handling through opcode 1821.
  - Validates that the requested record:
      - Exists
      - Is a Title
      - Has been claimed

  - Stores the equipped Title on the selected character.
  - Publishes it to the character screen and banner.
  - Supports toggling a Title off using the native unequipped sentinel.
  - Refreshes the relevant Family 4 state immediately so equip and unequip work without leaving the menu.
  - Persists the equipped state through normal account persistence.

  ## Seasonal XP

  - Added persistent Seasonal XP state.
  - Activates the installed Season of Arrivals progression.
  - Uses progression definition 40 for cumulative Seasonal XP.
  - Uses progression definition 41 for the gameplay HUD bar.
  - Calculates ranks at 100,000 XP per rank, capped at rank 100.
  - Publishes XP simultaneously to:
      - The Seasons screen
      - Inventory progression state
      - The normal gameplay HUD bar

  - Generates the native +XP presentation when XP is awarded.
  - Migrates the earlier XP persistence format without losing accumulated XP.

  ## Season Pass rewards

  - Added Season Pass reward claiming through opcode 2400.
  - Mapped all 196 Season of Arrivals reward rows.
  - Validates progression, reward index, required rank, item definition, and claimed state.
  - Supports currencies, consumables, cosmetics, weapons, armour, and class-specific rewards.
  - Expands the rank-one premium package into the correct class armour set and weapon.
  - Persists claimed rewards independently from XP.
  - Publishes claimed flags into the account unlock bank.
  - Changes claimed tiles to the native unavailable appearance with the Already claimed tooltip.
  - Prevents duplicate reward grants.
  - Sends granted items through the Queuez acquisition feed.

  ## Queuez acquisition banner

  - Added a deferred world-reward queue for collectible and progression rewards.
  - Supports multiple pending rewards without one overwriting another.
  - Keeps acquisition presentation independent from the live inventory slot contents.
  - Prevents equipping a newly acquired item from changing the banner to the displaced item.
  - Preserves all active acquisition rows when several items are displayed.
  - Holds presentation state for the full eight-second banner lifetime.
  - Safely expires presentation state without causing a later porpoise disconnect.
  - Rewards fall back to direct commit after bounded publication retries rather than remaining permanently queued.

  ## Build-data and persistence work

  - Added extracted catalogs for:
      - Records
      - Record rewards
      - Presentation nodes
      - SObjects
      - Collectibles

  - Extended the build-data cache format for the new domains.
  - Added cache validation and automatic rebuild support.
  - Consolidated node and record persistence into the shared build-data cache.
  - Added atomic persistence for claims, claimability, objective progress, Seasonal XP, and Season Pass claims.
  - Invalid or outdated cache formats are rejected and rebuilt safely.

  ## Cleanup and restructuring

  The final cleanup pass:

  - Removed obsolete per-domain record and node persistence implementations.
  - Removed the abandoned bubble-based lore mapping table.
  - Removed unused activity-state and incident-encoder code.
  - Consolidated repeated Family 4 snapshot and reward publication logic.
  - Simplified record reward, Season Pass reward, equipment, and inventory transaction handling.
  - Replaced duplicated objective lookup logic with compact generated tables.
  - Flattened nested control flow and standardized naming.
  - Reduced noisy logging and legacy fallback paths.
  - Preserved transactional preview/commit behavior for inventory mutations.
  - Reduced the cleanup commit by approximately 5,800 net lines while retaining the implemented behavior.

  ## Verified behavior

  The following have been tested in-game:

  - Triumph claiming, score, category counting, and parent bars
  - Lore chapter visibility and claiming
  - Counter-granted lore books
  - Exact Dreaming City collectible mapping
  - Dead Ghost duplicate protection
  - Fallen drones
  - Menagerie vases
  - Dust collectibles
  - Title equip, immediate refresh, and unequip
  - Corrupted Egg Dreaming City rewards
  - Cat statue rewards and Triumph progress
  - Jade Rabbit Moon rewards and Triumph progress
  - Moon Lost Ghost rewards and Seasonal XP
  - Seasonal XP on the Seasons screen and gameplay HUD
  - Season Pass reward grants, claimed presentation, and duplicate prevention
  - Multiple Queuez acquisitions and immediate equipment swaps

  ## Current limitations

  - Necrotic Ciphers are not implemented due to the current account state having them all marked as collected. I will resolve this at a later time, this PR is already big.
  - World spawning/visibility of missing Corrupted Eggs is outside this PR; the server handles their incidents when they occur.
  - Collectible item rewards are currently repeatable for testing, while lore and Triumph progression remain deduplicated.
  - Seasonal progression is authored specifically for the installed Season of Arrivals build.

Nyxaraa:
## Changes since the last push

  Range: 4e07fe5 → ced30dd

  ### Seasonal Artifact

  - Added the Seed of Silver Wings to the correct Artifact equipment slot for every character.
  - Enabled the Season of Arrivals Artifact without creating a second invisible Artifact item or repeatedly triggering the Arrivals intro.
  - Added server handling for Artifact perk purchases through opcode 901.
  - Implemented:
      - Available perk points
      - Spent perk count
      - Column unlock requirements
      - Persistent perk ownership
      - Immediate Artifact UI refresh

  - Artifact mods now behave as native socketable mods.
  - Artifact purchases appear in the acquisition feed.
  - Artifact purchase state persists between restarts.

  ### Reset Artifact

  - Implemented the Reset Artifact action.
  - Deducts the advertised 20,000 Glimmer.
  - Clears all purchased Artifact perks.
  - Refunds all spent perk points.
  - Removes affected Artifact mods from equipped and inventory item sockets.
  - Restores each socket’s manifest-defined initial plug.
  - Refreshes the Artifact screen immediately.
  - Suppresses the unwanted reset entry from the item acquisition feed.
  - Refreshes Family 0/3 derived perk data after reset, ensuring champion effects stop immediately rather than persisting until another mod
    is applied.

  ### Artifact XP and Bonus Power

  - Connected Artifact progression to earned Seasonal XP.
  - Added the native escalating XP curve for Artifact bonus Power.
  - Projects:
      - Artifact Power bonus
      - Earned perk points
      - Spent perk points
      - Purchased perk flags

  - Bonus Power now appears separately as the blue +N Bonus value.
  - Bonus Power contributes to the character’s effective Power without incorrectly raising the underlying gear score.
  - Season rank and HUD XP now use their correct separate progression rows:
      - Slot 40: cumulative Season Pass progress
      - Slot 41: current-rank HUD progress

  - Rank 100 correctly caps the pass while allowing post-cap XP to continue.
  - No authored Season Rank or test rank is included in the committed settings.

  ### Artifact persistence

  - Upgraded the Seasonal XP cache format to persist the Artifact purchase mask.
  - Existing version-one and version-two XP/claim files remain loadable.
  - Writes remain transactional through a temporary file and atomic replacement.
  - Artifact mutations use stale-state checks to avoid overwriting concurrent changes.

  ### Live progression refresh

  - Added live Family-5 override publication and investment rebuilding.
  - Artifact points, ownership and Power values update without restarting the game.
  - Added selected-character Family-4 refreshes for purchases and resets.
  - Reset-affected item residents are republished individually to avoid full-inventory acquisition behaviour and Queuez instability.

  ### Season Pass claiming

  - Added the complete Season of Arrivals reward catalogue: 196 native reward rows.
  - Opcode 2400 now validates:
      - Correct progression
      - Reward index
      - Required rank
      - Existing claimed state
      - Installed item definition

  - Claimed rewards are persisted and cannot be claimed twice.
  - Tiles immediately change to the native locked-style state with the Already claimed tooltip.
  - Claimed flags are mapped to their exact account flag rows, including the native gap in the flag sequence.
  - Claims and rewards commit together only after the response and Queuez update have been staged successfully.

  ### Season Pass rewards

  Implemented support for the different reward forms used by the pass:

  - Normal weapons, armour, currencies and materials
  - Rank-one premium class packages
      - Correct class armour set
      - Seasonal weapon

  - Destination Resource Bundles
      - Grants all nine destination materials
      - 50 of each material

  - Legendary Engrams
      - Automatically decrypt into a random eligible Legendary weapon or class-appropriate armour piece

  - Exotic Engrams
      - Automatically decrypt into a random eligible Exotic weapon or class-appropriate armour piece

  - Mnemonic and Umbral Engrams
      - Enter the character’s engram inventory instead of auto-decrypting

  - Added support for character inventory buckets without an equipment slot, required by stored engrams.
  - Season Pass reward grants use the normal Queuez acquisition feed.

  ### Exact Season Pass armour rolls

  - Added all 36 advertised armour rolls:
      - Hunter, Titan and Warlock
      - Helmet, arms, chest and legs
      - Three pass variants for each slot

  - Each reward index maps to its exact item and stat-plug combination.
  - Added compile-time validation to prevent a roll being assigned to the wrong reward item.
  - Class items are excluded from stat-roll authoring.

  ### Transaction and inventory handling

  - Expanded prepared reward batches to support the nine-item destination package.
  - Added atomic direct-item bundle handling for premium packages.
  - Added validation for auto-decrypted engram results.
  - Added Queuez staging for Season Pass item, profile-stack, package and resource-batch rewards.
  - Preserved normal item identities, mutation serials and inventory positioning.

  ### Cleanup pass

  - Removed the obsolete no-op Artifact refresh API and its callers.
  - Consolidated duplicated Season XP, maximum-rank, Artifact identity and sale-count constants.
  - Centralized class-specific engram pool selection.
  - Centralized selected-character resolution in deferred Artifact updates.
  - Extracted shared Family-4 character compression/commit logic used by equipment, item-state and Artifact updates.
  - Split Artifact reset socket/profile operations into focused helpers.
  - Added protection against dereferencing a missing prior equipped item during reset.
  - Cleanup reduced the affected code by a net 16 lines while retaining tested behaviour.

  ### Verification completed

  - Full clean build passes.
  - Equipment and item-state actions work.
  - Seasonal XP, rank display and Artifact progression work.
  - Artifact purchases refresh and persist.
  - Artifact reset correctly charges, clears and refreshes.
  - Champion effects disappear immediately after reset.
  - Season Pass claims show as claimed.
  - Destination bundles, engrams and exact armour rolls work.
  - Multiple Queuez acquisition entries remain stable without Porpoise disconnects.

Nyxaraa:
### Upstream merge and compatibility

  Merged the latest upstream build into triumph-work-code, retaining upstream’s newer mission-scripting, activity, networking, and
  character-state systems while restoring the branch’s progression features. Mission scripting and Activity SDK generation are enabled in
  the bundled defaults.

  ### Artifact mods and socket menus

  Fixed Flourishing Blade, Automatic Prize, Dimensional Tithes, and Ascendant Bounty appearing in General Armor Mod sockets instead of Leg
  Armor Mod sockets.

  The repair corrects both native menu membership and plug categories. It also preserves each mod’s own unlock conditions. Earlier
  approaches broke other mods because copying socket rows without relocating their nested relative pointers corrupted those conditions.

  The four mods now:

  - Appear in the Leg Armor Mod socket and can be equipped.
  - Appear immediately after Empty Mod Socket.
  - Leave existing general and leg mods available.

  The correction runs at content-table initialization, with bounded readiness retries rather than continuous scanning.

  ### Triumph claims and rewards

  Restored record claiming and progress synchronization, keeping partial progress, claimable, and claimed states separate.

  - Added proper multi-step triumph redemption tracking, including persisted redeemed-step counts and per-step score.
  - Fixed the MMXIX Reward: Sparrow and Reward: Emblem records by satisfying the missing claim-enabling conditions. These cards previously
    appeared claimable but never sent a claim request.

  - Updated client-derived investment state after committed account changes so claim availability and displayed progress refresh correctly.
  - Restored record reward handling and the progression needed by lore-book parent triumphs and titles.

  ### Collectibles and immediate rewards

  Restored lore collectibles, small cats, Jade Rabbits, and corrupted eggs to the collection/progression pipeline.

  Collection events now update the relevant lore chapters, triumph progress, and rewards. Account resynchronization includes the active
  connection, fixing rewards that previously appeared only after leaving the destination. Appearance-refresh failures no longer hold back
  otherwise valid progression updates.

  ### Titles

  Restored title application to the character after equipping.

  The title-selection logic was intact, but the appearance-refresh validator still expected the older raw deletion marker. Upstream now
  uses a codec-free none marker. Updating the validator allows the character refresh to reach the client, so equipped titles actually
  appear.

  ### Spawning, transitions, and fade-in

  Removed the manual fade-release workaround and its configuration option. It interfered with the native world-entry sequence, repeatedly
  triggered during public-area transitions, and disrupted the activity/bubble/closest-spawn display.

  Restored the native in-world spawn boundary and removed destination-specific workarounds, including Trostland-specific handling. Added
  throttled spawn-refusal diagnostics to make remaining destination-loading problems traceable without introducing further forced-spawn or
  fade overrides.

  ### Season pass

  Added Arrivals season identification, seasonal progression, reward definitions, and reward-claim handling to address the redacted season
  name and missing reward track.

  ### Lore visibility

  Fixed fresh-state visibility for:

  - Eva’s Journey
  - The Chronicon
  - All Wall of Wishes child triumphs
  - Confessions chapters

  This is a visibility-only repair: it does not grant chapters, progress, rewards, or claims. It validates the exact native identities and
  expression shapes before changing 36 visibility-condition instructions, checks that their storage is not shared with unrelated entries,
  and retains originals for rollback.

  ### Cleanup and validation

  Updated build-project entries, removed superseded workarounds, and added portable regression tests for socket-row relocation and lore
  visibility. Builds and both regression tests pass. The final socket behavior, title application, and fresh-state lore visibility were
  confirmed in-game.

  Personal claim resets used during testing are local save changes and are not included in the PR.

stanuwu:
Merged, thanks

PR #85, opened 2026-08-21
Fix: Exotic Catalyst investment state

yazan-albaiz:
# Description

## Issue

Sunrise did not send all state that Destiny 2 needs for completed Exotic weapon catalysts.

It rejected the Masterwork item flag `0x4` in some item-state checks. It also sent only part of the Family 5 completion state.

Some catalyst perks could work while the item screen still showed an incomplete catalyst. Worldline Zero could lose its catalyst socket or gold Masterwork border.

The Masterwork flag alone does not complete the item screen. The client also needs these items:

- The active effect plug.
- All required Family 5 flags and values.
- Each account objective that the progress plug references.

## Fix

This change:

- Accepts `Locked`, `Tracked`, and `Masterwork` item flags through mask `0x7`.
- Rejects item-state bits outside mask `0x7`.
- Reads released catalyst relations from the installed investment data.
- Limits released catalysts to the target Season 11 build.
- Supports the older three-plug layout and the newer two-plug layout.
- Sets the active effect plug and Masterwork flag in one checked item update.
- Keeps the existing Locked and Tracked flags.
- Sends each required Family 5 acquisition flag, completion flag, and completion value.
- Completes each account objective that a catalyst progress plug references.
- Leaves unreleased placeholder catalyst sockets unchanged.
- Stores the new catalyst data in build cache format 51.
- Adds `complete_exotic_catalysts`, which is enabled by default.
- Does not change the settings version.

## Tests Ran

- [x] Built Debug and Release with no warnings or errors.
- [x] Ran the local regression tests in Debug and Release.
- [x] Tested a new build cache and a reused build cache.
- [x] Ran latest commit in the installed ProjectSunrise Destiny 2 client.
- [x] Confirmed 45 released catalyst records and 30 placeholder records.
- [x] Confirmed zero unsupported catalyst records.
- [x] Tested and Confirmed that Worldline Zero, Witherhoard and Tractor Cannon showed their catalyst socket, gold Masterwork border and are fully functional.
- [x] Ran all client tests only in the installed ProjectSunrise installation.

## Credit

"jane bond" on the Sunrise discord server for pointing out that functionality was working pre 0.3.0

image


stanuwu:
Merged, thanks

PR #87, opened 2026-08-24
Tower vendors: purchases, quests, bounties and recycling

chnsw:
Tower vendors answer their interactions. Four behaviours, all server-side, driven by authored rule files that are re-read on use, so tuning them needs no rebuild:

- **Purchases** (opcode 901) — a sale row resolves to its item and grants through the same acquisition path a Collections pull uses. Cost is deliberately not charged: the row's cost-bearing fields are still role-open.
- **Quests** (opcode 904, new codec) — grant through the same path. A placeholder row grants what it stands for rather than the Dummy the window draws, and the answered banner retires so the next interaction can reach the screen.
- **Bounties** — an "Additional Bounties" row rolls one random unheld repeatable bounty from an authored pool, capped at five held as retail allows. Exercised in-game on Lord Shaxx and the Drifter; rules also ship for Banshee-44, Zavala, Eva Levante (all four events), the Prismatic Recaster and Saint-14, authored from the manifest and checked against it offline but not yet clicked in game.
- **Recycling** — the Drifter's synth rows and Master Rahool's 277 shader rows charge the stack they name and credit the payout, announced through the account change ring so the on-screen "+N" gain draws and accumulates.

Supporting fixes: vendor definitions are read for vendors named by hash rather than only the head of the index (the Drifter is row 195, so nothing past the head ever resolved), pursuits can be placed, held and discarded, and a dismantle no longer reshuffles the bucket's order.

Every refusal path is fail-closed, on the rule that a wrong grant which commits cleanly is worse than a refusal: a matched rule that cannot be honoured — an exchange with too many payouts, a substitution whose replacement is not installed — owns its row, grants nothing and logs what it got wrong instead of falling through to the placeholder; a 904 body without the 32-bit sale-row field is refused rather than guessed at; and the banner retires only when its offer was actually answered, so a transient grant failure no longer buries a quest offer for the session. The vendor definition pass skips a definition that cannot fit instead of dropping the whole catalog, and a cached boot that carries no catalog retries the pass once instead of never.

The cache format version moves to 46: the cache stores the vendor domains, an older cache would silently restore the old set, and the pass now publishes partial catalogs where it published all or nothing.

Driven by four rule files installed to `bin\x64\Sunrise\` beside `settings.json`. Reference copies ship in the repo under `Sunrise/resources/vendor_rules`, exactly as run for the verification, with a note on where they install — so merging the code alone ships a working feature rather than one waiting on files from a conversation.

Verified in-game on build 86657: purchases, quests with banner retirement, repeatable rolls on Shaxx and the Drifter, both recycles with the gain toast, and pursuit discard — re-run whole after the refusal-hardening rework. The branch builds clean at /W4 with warnings-as-errors and boots with the full catalog published (48 definitions, 5,347 sale rows, zero skips). Every commit builds clean, so the branch bisects.


chnsw:
The rule files now ship in the repo under `Sunrise/resources/vendor_rules`, exactly as run for the verification, with a README on where they install (`bind\Sunrise\` beside `settings.json`).

chnsw:
The branch has been rewritten since the last push, so the history looks different: the twenty commits that grew this feature, several of which added something and later removed it, are now seven that each stand alone and each build clean at /W4 with warnings-as-errors. Every behaviour in the description was re-verified in game after each round of changes below.

**Fixes**

- Skipping a vendor definition that does not fit no longer empties the catalog. Sale rows were appended before the installed rows were tried, so a definition whose sale rows fit but whose installed rows did not left orphan rows in the bank; the next definition's offset carried the gap and `valid()` rejected the whole set, which was the fail-whole outcome the skip existed to remove. A skipped definition now leaves both banks as it found them.
- The vendor banner is answered when its grant commits, not when it is prepared. Preflight or the commit's staleness guard can still drop a prepared grant, and the answered list is append-only for the session, so a quest the player was still owed could have had its banner buried until relaunch. The answering vendor rides the acquisition transaction and is written in `service_outcome_commit`.

**Structure**

- The answered-interaction list lives in `state::vendors`. The client hook on the picker's retire test reads it and records what each vendor is showing; the server writes it. The server no longer includes a client hook, so its only client include is the network consumer, as on master.
- The vendor catalog is built in one pass. Definition selection moved into `vendors::build`, which already holds the index, replacing the publish, snapshot, clear, publish sequence.
- Vendor rows and collectibles are read through the catalogs under their own locks (`sale_row`, `installed_row`, `find_granting`) instead of copying whole tables into static buffers. That is 2 MiB off the DLL on disk and about 3 MiB of static storage.
- `core::log::writef` replaces the hand-rolled array, `snprintf`, guard and `write` at each new log site. The 901 and 904 codecs share one biased-index reader. The 901 clock rule, which had lost its only caller, is gone; the clock is still decoded so the body is checked whole.
- The sale row's +100 is named `categoryIndex`, which correlating 3,304 rows against the manifest settled it to be. The cache record's field is renamed with type and position unchanged, so the on-disk format is not.
- Comments are trimmed toward the density of the rest of the tree. History that belongs in the project notes is out; the reasoning stays.

**Rule files**

- Master Rahool and Amanda Holliday are named in `vendor_catalog.txt`. The exchange and substitution files were already theirs, and both vendors were only resolving because head-fill happened to reach them.

**Known and left alone**

- `opcode_routes.cpp` lists 901 as a status pair with a trailing boolean, while the mutating reply path re-encodes a successful purchase as a plain status pair, and the plain form is what the game accepts. The disagreement is recorded in a comment rather than corrected blind; only a captured retail reply settles it.

Footprint against master: 51 files, +2,870 / -332. Three files are new (`state/vendors/answered_interactions.{h,cpp}`, `messages/biased_field.h`); the rest of the widening is small touches in `core/logging`, the two catalogs, and the three BAP files that carry the banner answer.

stanuwu:
Merged, thanks

PR #88, opened 2026-08-24
Persist account state with SQLite

Confetti3:
## Summary

- Persist the existing account, character, equipment, inventory, socket, profile-item, reward, and settings state in SQLite schema v1.
- Restore a validated snapshot during startup, with `settings.json` retained as bootstrap and recovery input.
- Save the complete account atomically during shutdown using the Windows SDK WinSQLite API.

## Implementation

The database lives at `Sunrise/state.sqlite3` beside the existing generated artifacts. Relational account data is normalized across six tables; the fixed settings leaf uses an explicitly encoded, versioned payload rather than copying C++ object bytes.

Startup validates the schema, account-row format, settings-payload format, bounded row counts, and the reconstructed production `AccountState`. A missing or malformed snapshot falls back to `settings.json`. Newer schema or payload formats also fall back, but disable persistence writes for that session so older code cannot overwrite newer data.

Shutdown writes a durable checkpoint before Client teardown. If that checkpoint fails, shutdown returns false while Client, Server, Middleware, and State remain alive for retry. The final post-teardown save is best-effort because the same account image is already durable.

## Compatibility

- Schema remains version 1.
- No public API or settings-format changes.
- `settings.json` remains unchanged and recoverable.
- Newer schemas are inspected before writable pragmas and left untouched.
- Existing account settings and key-binding source are preserved; live client settings propagation is outside this PR.
- WinSQLite is supplied by the declared Windows SDK; no SQLite binary is vendored.

## Validation

- Clean `Release|x64` rebuild with Visual Studio 2026 Build Tools, toolset `v145`, and Windows SDK `10.0.26100.0`.
- Standalone production-validator integration test: `sqlite-account-snapshot-ok`.
- Complete field-by-field account round trip, including high-bit identifiers and nullable socket plugs.
- `PRAGMA integrity_check` and `foreign_key_check` pass.
- Invalid production state is rejected.
- Newer schema, account-row, and oversized settings-payload formats remain unchanged across initialize/shutdown.
- Newer schema in DELETE journal mode remains in DELETE mode after inspection.
- Branch-wide `git diff --check` and 100-column check pass.
- Manual restart testing covered character selection, equipment changes, socket selection, item lock state, profile acquisition, dismantling rewards, and restored snapshots.

## Known limitation

Persistence currently checkpoints on clean shutdown. A forced termination can lose changes made after the previous clean snapshot, while the last committed database remains valid. Live audio and key-binding mutations are not propagated back into account State by this PR.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -74,5 +74,5 @@ ExtraArgsBefore:
   - -Wdocumentation
 FormatStyle: file
 CheckOptions:
-  portability-restrict-system-includes.Includes: '-*,Windows.h,WinSock2.h,WS2tcpip.h,MSWSock.h,WinDNS.h,TlHelp32.h,Shellapi.h,bcrypt.h,d3d11.h,detours.h,dxgi.h,wincodec.h,imgui.h,imgui_impl_dx11.h,imgui_impl_win32.h,intrin.h,algorithm,array,atomic,bit,bitset,cctype,charconv,chrono,climits,cmath,cstdarg,cstddef,cstdint,cstdio,cstdlib,cstring,cwchar,limits,memory,new,optional,span,string_view,type_traits,utility,variant,vector'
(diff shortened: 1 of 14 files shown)


ocucor (approved):
Clean integration, shouldn't interfere with other PRs. LGTM 👍 🔥

stanuwu:
already said i will do this later and this is not the right time
not PR features

PR #89, opened 2026-08-24
ci: build the Linux (clang-cl/xwin) cross-compile on every PR

ProjectUgarte:
The Linux build path has broken twice since it landed (#40, #61) because CI only exercises MSVC. This adds a second job that runs the README's documented Linux build on `ubuntu-24.04`, so a PR that compiles under MSVC but not the cross toolchain fails checks instead of landing.

Details:
- LLVM 20 from apt.llvm.org — the MSVC STL that xwin fetches static-asserts Clang >= 19 (stock ubuntu-24.04 clang-18 fails), and the project's warning flags need >= 19 as well
- xwin 0.10.0 pinned; the splatted SDK is cached between runs (one-time ~2 min fetch on first run)
- Ninja + the repo's own `linux-to-win-toolchain.cmake`, per README; artifact uploaded from `build/x64/Release/`, mirroring the MSVC job
- Timing on my fork: 5m16s with warm cache — finishes before the MSVC job, so no added wall-clock
- Green run: https://github.com/ProjectUgarte/Sunrise/actions/runs/32770256080

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,8 +83,9 @@ jobs:
         run: cmake --build build
 
       - name: Upload artifact
+        # CMakeLists routes outputs to build/x64/<CONFIG>, mirroring MSVC.
         uses: actions/upload-artifact@v4
         with:
           name: Sunrise-Release-x64-linux-cross
-          path: build/steam_api64.dll
+          path: build/x64/Release/steam_api64.dll
           if-no-files-found: error
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,12 +45,16 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install LLVM toolchain and Ninja
-        # The toolchain file invokes the unsuffixed clang-cl / lld-link /
-        # llvm-lib / llvm-rc names, which Ubuntu only ships inside the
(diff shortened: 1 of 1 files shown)


TotalTaxAmount:
I do think that this is a good idea, not sure if this needs to upload cause they are both windows binary's, but I guess it doesn't hurt. Also any reason why you are manually downloading xwin instead of just using `cargo install xwin`?

ProjectUgarte:
Thanks! On the artifact: kept it deliberately — it lets anyone grab a hardware-testable DLL from any PR without a local toolchain (I'm testing PRs on a Steam Deck/Steam Machine, so a downloadable Linux-built artifact is handy for that), and it doubles as proof the link step produced a complete DLL rather than just objects. Happy to drop it if you'd rather keep CI lean, since the MSVC job already uploads one.

On xwin: purely CI speed/determinism — `cargo install xwin` compiles it from source on every uncached run (several minutes plus a Rust toolchain on the runner), while the pinned release tarball is a ~2s download and keeps the job's dependency surface to just LLVM. Same tool, same behavior. Could add the release's sha256 check if you want belt-and-suspenders.

stanuwu:
slop

PR #90, opened 2026-08-24
 Reduce boot time in the signature sweep and hook attach

SkyzerFlyzer:
Boot spends almost all of activate stage=main in two places: the pattern sweep that resolves signatures, and the Detours transactions that attach the boot-step fixes. This reduces both. No behaviour changes, no header changes, and no new settings.

**Measured**

  Two DLLs built from the same toolchain (v145 / SDK 10.0.26100.0), each hash-verified against the run that produced its log, booted to orbit on the same machine:
```
  ┌──────────────────────────────────────────┬──────────┬─────────────┐
  │                                          │ upstream │ this branch │
  ├──────────────────────────────────────────┼──────────┼─────────────┤
  │ activate stage=main                      │ 3172 ms  │ 328 ms      │
  ├──────────────────────────────────────────┼──────────┼─────────────┤
  │ bootflow attach span (first fix to last) │ 1000 ms  │ 16 ms       │
  ├──────────────────────────────────────────┼──────────┼─────────────┤
  │ initialize                               │ 140 ms   │ 125 ms      │
  └──────────────────────────────────────────┴──────────┴─────────────┘
```
  initialize is unchanged as expected — nothing here touches that phase. Run-to-run variance on activate stage=main is about two 15.6 ms ticks, far below the difference.

  Functionally identical across both runs: 11 bootflow stages ok, 39 hooks attached, 31 egress hooks installed, zero failures and zero signature misses on each side.

**What changed**

  Anchor pattern scans on the rarest exact byte, not the first.
  anchor_of took the first exact byte in a signature. For an x64 function prologue that is almost always a REX prefix, and 0x48 alone is a large share of the bytes in compiled x64 code, so memchr stopped to verify constantly across the sweep's whole-image passes. It now
  picks the least frequent exact byte, measured against the ranges about to be scanned rather than guessed. A 256-bin histogram of those ranges is built once and cached, keyed on a fingerprint of the range pointers and sizes — without the cache every pattern would pay
  the extra traversal the anchor choice exists to avoid. Ties go to the earliest byte, so one image always picks the same anchor.

  Attach the boot-step fixes in one transaction.
  bootflow::install ran eleven fixes, nine of which each opened their own Detours transaction to attach a single detour. A transaction costs far more than the attach it guards: it enlists the threads it must suspend, and enlisting walks every thread on the system twice.
  Each of the nine is split into a stage that resolves its target and fills a Spec, and a publish that takes the resulting handle, stores the trampoline and reports. world_step and fade_release stay outside — they only find addresses to call and open no transaction.

  Enlist transaction threads without a system-wide snapshot.
  Enlisting walked a Toolhelp snapshot, which enumerates every thread on the system to reach this process's own. It now walks the process's own thread list through ntdll instead.

**Correctness**

  The anchor change is match-set-preserving by construction: matches_at still verifies every exact byte, and next_candidate already shifted its search window by the anchor index and subtracted it back off the hit, so an anchor anywhere in the pattern was always
  supported. A differential harness comparing the old and new selection across many patterns and seeds — covering unique, missing, ambiguous and invalid results, plus anchors at the head and tail of a pattern — reports identical matches for every one. The identical hook
  and signature counts in the table above are independent confirmation on a real image.

  Batching preserves the previous failure isolation. A fix whose target is missing is simply not in the batch, so one miss cannot cost the others their fix; if the batch itself fails, the fixes are retried one at a time, so a single target Detours refuses cannot take
  the group down. The stage result is a tri-state rather than a bool, because a missing target and an already-attached fix both stage nothing but mean opposite things for whether the group succeeded.

**Review notes**

  The ntdll thread walk is the part most worth a close look.

  The export is undocumented, so a build without it keeps the snapshot, which is also the fallback when the walk stops early. The walk reaches threads a snapshot never reports: a thread that has exited still sits on the kernel list for as long as anything holds a handle
  to it, and this process leaks two. That distinction matters more than it looks — DetourUpdateThread suspends a thread the moment it is handed over, suspending an exited thread fails, and Detours records that as a transaction-wide pending error which DetourAttach
  returns early and commit aborts on. Nothing can clear it, so one exited thread would cost every hook in the transaction. Threads are therefore checked with GetExitCodeThread before they are offered. The enlist handle still comes from OpenThread with the access the
  snapshot pass used, so the set of threads a transaction holds is unchanged apart from the exited ones it should never have held.
  
**Human Message**

Utilized Claude Code to do most of this. Performed manual testing and was unable to find any regressions in loading. Attached some videos of before and after

https://github.com/user-attachments/assets/3d27bd8e-03b6-471b-ac9f-6d1fc4a77305

https://github.com/user-attachments/assets/0fd78814-a80e-49db-a826-0e2085ed9ed2

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -1057,6 +1057,8 @@
     <ClCompile Include="src\client\content\items\packages\package_build_report.cpp" />
     <ClCompile Include="src\client\content\items\packages\package_root_tables.cpp" />
     <ClCompile Include="src\client\content\items\packages\package_item_rows.cpp" />
+    <ClCompile Include="src\client\diagnostics\entity_create_probe.cpp" />
+    <ClCompile Include="src\client\diagnostics\image_dump.cpp" />
     <ClCompile Include="src\client\diagnostics\module_range.cpp" />
     <ClCompile Include="src\client\process\freeze\client_process_freeze.cpp" />
     <ClCompile Include="src\core\settings\address_text.cpp" />
@@ -1866,6 +1868,8 @@
     <ClInclude Include="src\server\transport\internal.h" />
     <ClInclude Include="src\middleware\content\packages\reader\locator_cache.h" />
     <ClInclude Include="src\server\bap\encrypted\push\queuez\queuez_push_reporting.h" />
+    <ClInclude Include="src\client\diagnostics\entity_create_probe.h" />
+    <ClInclude Include="src\client\diagnostics\image_dump.h" />
     <ClInclude Include="src\client\diagnostics\module_range.h" />
     <ClInclude Include="src\client\process\freeze\client_process_freeze.h" />
     <ClInclude Include="src\core\settings\address_text.h" />
--- a/Sunrise/resources/default_settings.json
+++ b/Sunrise/resources/default_settings.json
@@ -50,7 +50,8 @@
       "transport_address": "127.0.0.1",
       "port": 30976,
       "server_reserve_count": 256,
-      "client_join_grant_count": 8192
+      "client_join_grant_count": 8192,
+      "client_lease_high_water": 0
     },
     "activation": {
       "default_client_activation": true,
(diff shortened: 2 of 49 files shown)


stanuwu:
Merged, thanks

PR #91, opened 2026-08-25
Threading improvements

Polkm105:
## Summary

This is mostly a temp check PR to see if there's any appetite for this type of work on the back-end. Makes it harder to mess up threaded data access and improves semantics around `SRWLOCK`. Breakdown of the changes:

### Adds `SrwLock` class
A wrapper around SRWLOCK that complies with STL `Lockable` and `SharedLockable` concepts. Allows the use of stl locks like `std::lock_guard` and `std::shared_lock`. 
* Replaces instances of SRWLOCK + manual locking/unlocking with `SrwLock` + `std::lock_guard` where appropriate
* Removes `ExclusiveLock` from `server/gameplay/physics/world/command_queue.cpp`, replaced with `std::lock_guard` + `SrwLock`
* Removes `Lock` from `state/build_data/table.h`, replaced with `std::lock_guard` and `std::shared_lock` + `SrwLock`

### Adds `Sendable` concept
An indicator that a type can be safely copied across the thread boundary. This is used in tandem with `DataMutex` to improve semantics and prevent accidental ref leaks when working with data that's accessed by multiple threads.

Currently supports the following types:
* Anything satisfying `std::integral`
* Anything satisfying `std::floating_point`
* `void`

`struct IsSendable<T>` can be specialized to mark custom types as safely `Sendable`. This isn't a perfect solution but at least lowers the surface area from "accidental ref leak" bug to "intentional but wrong" bug

### Adds `DataMutex` and `SharedDataMutex` classes
Essentially `Data` + `Mutex` wrappers. This makes it impossible to read/write threaded data without the associated mutex being locked.

* `Data` types can be written without threading in mind as they'll only be accessible when the associated mutex is locked
* Writes during a shared lock are impossible as you're given a `const Data&`
* Reduces the risk of accidental pointer/ref leaks of threaded data by only allowing the return of `Sendable` types

## Verification
* No additional warnings/error from `cmake` builds in both Debug and Release 
* `clang-format --style=file --dry-run --Werror` reports no issues
* Confirmed matching existing behavior through debugging

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -74,5 +74,5 @@ ExtraArgsBefore:
   - -Wdocumentation
 FormatStyle: file
 CheckOptions:
-  portability-restrict-system-includes.Includes: '-*,Windows.h,WinSock2.h,WS2tcpip.h,MSWSock.h,WinDNS.h,TlHelp32.h,Shellapi.h,bcrypt.h,d3d11.h,detours.h,dxgi.h,wincodec.h,imgui.h,imgui_impl_dx11.h,imgui_impl_win32.h,intrin.h,algorithm,array,atomic,bit,bitset,cctype,charconv,chrono,climits,cmath,cstdarg,cstddef,cstdint,cstdio,cstdlib,cstring,cwchar,limits,memory,new,optional,span,string_view,type_traits,utility,variant,vector'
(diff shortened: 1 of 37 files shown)


stanuwu:
Merged, thanks

PR #93, opened 2026-08-30
Investement: random rolls - support randomized weapon acquisitions

echo-matt:
## Summary

Adds server-authored random rolls for eligible non-exotic weapons acquired into a character inventory.

- Separates each socket lane's curated offer set from its native-order randomized draw pool.
- Authors rolled perk plugs, weapon-valid Tier-1 masterworks, random-roll bytes, and per-instance owned-row masks.
- Keeps rolled plugs reachable after curated socket swaps through one shared offer rule.
- Corrects hover rendering after the native socket-mask rebuild so hover and inspection show the same rolled perk.
- Leaves armour, cosmetics, profile-owned rows, exotics, and shader/ornament/mod lanes on their authored definitions.

## Socket relations

A lane's embedded, reusable, and initial plugs form its small curated insertion set. The randomized set is stored separately in native row order and is used only as the draw pool. An instance publishes the randomized rows it owns; the rolled pick's bit is always retained.

Trait columns own exactly the rolled trait. Other rolled lanes own the rolled pick plus curated members that share rows with the randomized set.

## Safety and validation

- Item-level refusal happens before mutation.
- Randomized pools wider than the 64 representable owned-row bits fail closed instead of truncating.
- Socket validation and staging share the same curated-plus-owned offer rule.
- Socket-entry state is not used to pin rolls; hover/inspection agreement comes from the authored plug and `availablePlugRows`.
- Cache format is bumped from 44 to 50, including invalidation of stale version-49 curated-pool extraction.
- Detailed roll telemetry is debug-level.

## Scope

This PR implements randomized item acquisition. It does not include the local Player-panel drop button used for live testing or the future engram/world-drop producer.

The current Collections UI only submits reacquirable fixed/non-randomized definitions, so this does not make blocked randomized Collection gear pullable.

## Validation

- Repeated local test acquisitions produced distinct rolls.
- Hover and inspection matched without socket-entry pinning.
- Curated socket swaps remained reversible.
- Native Visual Studio MSBuild/MSVC, Release x64, toolset v145, `/W4 /WX`.

## Notes

No copyrighted data is included. Plug sets, socket rows, and item stats are extracted from the installed runtime build data. The remaining literals are identifiers and native index constants consistent with existing project style.

### DIFF
--- a/Sunrise/Sunrise.vcxproj
+++ b/Sunrise/Sunrise.vcxproj
@@ -145,6 +145,7 @@
     <ClCompile Include="src\state\runtime\state_account_profile_runtime.cpp" />
     <ClCompile Include="src\state\runtime\state_account_socket_runtime.cpp" />
     <ClCompile Include="src\state\runtime\state_rolled_socket_plugs.cpp" />
+    <ClCompile Include="src\state\runtime\state_item_random_roll.cpp" />
     <ClCompile Include="src\state\runtime\state_account_item_action_runtime.cpp" />
     <ClCompile Include="src\core\ui\busy\ui_busy_overlay.cpp" />
     <ClCompile Include="src\core\ui\busy\ui_busy_state.cpp" />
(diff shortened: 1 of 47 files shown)


stanuwu:
not properly in line with retail behaviour

echo-matt:
> not properly in line with retail behaviour

Would you please elaborate on what needs fixing @stanuwu ?

PR #96, opened 2026-09-04
Fix missing weapon Power in character stats

Confetti3:
Firearms could deal very low PvE damage even when the inventory showed the correct Power. The generated weapon stat tables omitted the Power row that the client uses for damage calculations.

Populate that row with each equipped weapon's own Power, using its identifier from the installed investment constants and the existing item-level calculation. Preserve the other weapon stats and reject invalid values or tables that cannot fit. Bump the build-data cache version so existing installations rebuild the cached constants with the new field.

Verified with the Release x64 build, focused stat-generation tests, and real cache round-trip and rejection tests. The cleaned-up patch was retested against Leviathan Loyalists and confirmed to resolve the firearm damage issue. Live inspection also confirmed the populated weapon Power and its expected conversion in damage state.

Changed-line clang-tidy passes. The full check reports an existing `reader.h` include-allowlist error, also reproduced on unchanged upstream.


stanuwu:
Merged, thanks

PR #99, opened 2026-09-05
Fix catalyst-enabled sign-in exceeding native unlock override capacity

chnsw:
With the shipped `complete_exotic_catalysts: true`, the default investment snapshot needed 130 Family-5 flag overrides: 43 authored, 25 artifact and 62 catalyst. That exceeded Sunrise's previous 127-row storage, and the client's Family-5 arrays only hold 100 rows each. On `a57dc9a` opcode 503 fell back to a 7-byte echo and the client disconnected during investment sign-in after a message-argument bitstream decode error.

This keeps completion enabled and brings the default snapshot down to 90 flags:

- Artifact ownership stays in its native character acquired-flag bank, which already carries the purchase and reset path. The 25 duplicate Family-5 artifact flags are dropped, and stale authored copies are removed so they cannot mask the character projection.
- Catalyst completion flags that the package's account flag-map table maps to the account bank are set there. Fifteen flags move; acquisition gates and the unmapped completion flags stay in Family 5. An authored override for a mapped completion flag is still raised to set.
- `kUnlockOverrideCapacity` becomes 100, the client's real array size. The family-5 rearm hook previously copied 127 rows into the client object, which ran 108 bytes past the flag list.
- Artifact projection and the investment snapshot fail atomically, and the snapshot logs the reason when it does.
- The derived account mappings persist in cache format 61 (master took 60 for the entity position profiles), so older build-data caches rebuild.

Changes since the first revision: rebased on current master after the tests and docs folders were removed, so the test, fixture and markdown files are gone; the touched lines follow `.clang-format`; the new declarations carry full doc comments; the builder's mapping failure reports the item hash and lane like its neighbours; and the snapshot logs its failure reason. Release x64 build is clean.

Offline verification against the installed package's 11,923-row account mapping table and the 75-row catalyst catalog: 45 released catalysts, 15 mapped completion flags, successful cache re-derivation, and a complete default snapshot of 90 flags / 42 values.

This PR changes native state projection only. It contains no gameplay scripts, Tower Events changes, runtime/account files, or unrelated networking fixes.

### DIFF
--- a/Sunrise/src/client/content/items/packages/package_item_rows.cpp
+++ b/Sunrise/src/client/content/items/packages/package_item_rows.cpp
@@ -1,9 +1,13 @@
 #include <array>
+#include <cstring>
 #include <span>
+#include <vector>
 
+#include "../../../../middleware/content/packages/tables/definition_index_table.h"
 #include "../../../../state/build_data/items/catalysts/exotic_catalyst_builder.h"
 #include "../../../../state/build_data/items/details/item_detail_catalog.h"
 #include "../../../../state/build_data/runtime.h"
+#include "../../../../state/unlocks/definition.h"
 #include "internal.h"
 #include "package_socket_plug_build.h"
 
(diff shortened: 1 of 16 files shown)


leahanderson1 (approved):
LGTM, tested this out and it fixed the weasel error codes.

stanuwu:
reviewed
- remove random tests this contains game data and can not be merged
- remove random mds

chnsw:
Removed the tests, the fixture files and the markdown, and rebased on current master (cache format moves to 61 since master took 60 for the entity position profiles). The remaining changes are the native projection only: touched lines now follow .clang-format, the new declarations carry full doc comments, the builder's mapping failure reports the item hash and lane like its neighbours, and the investment snapshot logs its failure reason. Release x64 build is clean.

chnsw:
Re-tested on a 0.4.0.0 build carrying this branch together with #103 over a fresh 0.3.2 settings file: investment sign-in completes with catalyst completion enabled and the Tower loads to in_world with no errors in the log.

stanuwu:
Merged, thanks.

Issue #100, opened 2026-09-05
errors using UMU/Proton


leahanderson1:
any boots after building the SDK do not launch, giving exit code 5.
The log only contains one line:
`state level=info t=8 ev=account stage=identity primary=0x9EAA300100100100 characters=3`

game boots when using vanilla wine, though.

lutralutraq77:
Hi @leahanderson1 — could you try the diagnostic build from [[this successful Windows CI run](https://github.com/lutralutraq77/Sunrise-work/actions/runs/33990738907)](https://github.com/lutralutraq77/Sunrise-work/actions/runs/33990738907)? It adds startup/SDK logging to help locate the failure; it isn’t a confirmed fix.

Back up your existing DLL, settings and generated SDK, then follow the [[UMU/Proton logging instructions](https://github.com/lutralutraq77/Sunrise-work/blob/4d2404de6d4cd41283b789c0465b549a0eb5f65c/docs/umu-sdk-startup-diagnostics.md)](https://github.com/lutralutraq77/Sunrise-work/blob/4d2404de6d4cd41283b789c0465b549a0eb5f65c/docs/umu-sdk-startup-diagnostics.md), including debug logging and `PROTON_LOG=1`. Please reproduce once and attach:

* The full `Sunrise/logs/sunrise.log`, Proton log (including any exception/backtrace), and UMU/launcher terminal output from that run.
* Exact Sunrise build/commit, UMU and Proton versions, plus the vanilla Wine version that works.
* Your launch command/options and environment variables, including DLL overrides and prefix selection.
* Whether working vanilla Wine and failing UMU/Proton used the same Wine prefix, game files and generated SDK. If different, please describe the differences and include a working Wine log if possible.

Save logs before another launch overwrites them, and redact credentials/personal details; no game files or SDK dumps are needed. The account-identity line is normal, and exit code 5 alone doesn’t identify the underlying error. Thanks!

leahanderson1:
> Hi [@leahanderson1](https://github.com/leahanderson1) — could you try the diagnostic build from [[this successful Windows CI run](https://github.com/lutralutraq77/Sunrise-work/actions/runs/33990738907)](https://github.com/lutralutraq77/Sunrise-work/actions/runs/33990738907)? It adds startup/SDK logging to help locate the failure; it isn’t a confirmed fix.
> 
[...]

PR #101, opened 2026-09-05
Fix activity SDK publication under Wine

Rhodderz:
Sunrise has a requirement for its path to be an ordinary directoy. Windows is fine with this, Wine not so due to its Z drive being a reparse point so just fails.
This makes the final steps fail
CHange is keeping windows the same but for wine Sunrise should now accept the wine driver root while maintaining some reject for anything below.

Did port this (from memory) from a mod for iirc Skyrim that under wine would do simular things when downloading new maps

### DIFF
--- a/Sunrise/src/client/content/activity/activity_sdk_live_publication.cpp
+++ b/Sunrise/src/client/content/activity/activity_sdk_live_publication.cpp
@@ -69,33 +69,44 @@ struct Marker final {
     }
 }
 
-/** Requires every existing drive-path directory component to be ordinary, never a reparse point. */
-[[nodiscard]] bool ordinary_ancestry(const std::wstring& directory) noexcept {
+/** Names the first drive-path component which is not an ordinary directory. */
+[[nodiscard]] const char* ordinary_ancestry_reason(const std::wstring& directory) noexcept {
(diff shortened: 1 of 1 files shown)


leahanderson1:
this could (possibly?) fix #100

Rhodderz:
That looks like the same error I was having originally and made me look a little  closer.
Did see someone mentioning in chat about getting this working on mac and in theory it should work fine over there as i believe Wine does the exact same handling of its storage

leahanderson1:
> That looks like the same error I was having originally and made me look a little closer. Did see someone mentioning in chat about getting this working on mac and in theory it should work fine over there as i believe Wine does the exact same handling of its storage

just tried, this doesn't fix my issue

Rhodderz:
I never actually checked a relaunch
Had to move to GE-Proton11-3 which got it to work, now hitting weasle which is expected.

leahanderson1:
> I never actually checked a relaunch Had to move to GE-Proton11-3 which got it to work, now hitting weasle which is expected.

Switching to GE-Proton 11-3 works even without your PR.

stanuwu:
Merged, thanks.

PR #102, opened 2026-09-05
Fix empty settings documents during version migration

lutralutraq77:
This fixes a startup failure when upgrading a valid empty settings file.

For `{}`, the version upgrader currently produces `{"version": 13,}`. The trailing comma makes the generated JSON invalid, so settings parsing fails and startup stops.

In `Sunrise/src/core/settings/settings_upgrade.cpp`, the fix checks the first non-whitespace character after the opening brace. Empty objects receive the version without a comma; nonempty objects retain the separator and their original contents and whitespace.

The unchanged `kSettingsVersion = 13` constant moves into `Sunrise/src/core/settings/version.h`, allowing the production upgrader to compile independently of the Windows UI dependencies for testing. The Visual Studio project includes the new header.

`tests/settings_upgrade/` adds nine synthetic JSON regression cases covering empty objects, whitespace, nonempty objects, braces inside strings, and existing version fields. Each also checks exact-fit buffers, rejection of undersized buffers, and recognition of the upgraded version. A dedicated GitHub Actions job runs the tests.

Validation:

* All four empty-object cases failed before the fix; all nine cases passed afterward.
* The nine cases also passed with the fix applied to upstream `a57dc9a9e74eb71e876baea79047c9af679b2347`.
* CMake/CTest and address/undefined-behaviour sanitizer checks passed on the fork version.
* The equivalent fix on the older fork passed the Windows Release x64 build and regression tests: https://github.com/lutralutraq77/Sunrise-work/actions/runs/33977798758. That run does not validate the full Windows build of this upstream-based branch; this PR’s CI still needs to do that.
* In-game behaviour has not been tested.

This is one focused fix, with no new dependencies or game data and no settings-version increase.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,23 @@ on:
   workflow_dispatch:
 
 jobs:
+  settings-upgrade:
+    name: Settings upgrade regressions
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Configure tests
+        run: cmake -S tests/settings_upgrade -B build/settings-upgrade-tests -DCMAKE_BUILD_TYPE=Release
+
+      - name: Build tests
+        run: cmake --build build/settings-upgrade-tests
+
+      - name: Run tests
+        run: ctest --test-dir build/settings-upgrade-tests --output-on-failure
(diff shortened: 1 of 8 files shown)


stanuwu:
not planned, settings will be replaced soon

PR #103, opened 2026-09-05
Fix activity host connections and spawn ordering during travel

chnsw:
Direct travel between destinations can leave the screen black when the player spawns before the world transition finishes arming its fade. Keep the native `awaiting_client_sync` hold set until the client reports arrival through WS-702 world-state 8, while the loading lifetime still advances once the destination region is instantiated. That report follows the bootflow's arrival independently of the player spawn, so the hold cannot stall the transition, and the normal player spawn releases the fade after loading completes without a client patch or a fixed timeout.

Peer connect responses now advertise the host-pool port that received the request instead of always naming the primary gameplay port. The configured advertised IPv4 address is preserved, with the existing primary-port fallback when the receiving port is zero.

Upgrading from 0.3.2 could leave activity loading stuck when older settings omit `core.activity_sdk_generation.enabled`. The C++ default was false despite the bundled settings enabling generation, leaving the activity host unable to construct its roster without the SDK metadata. The missing setting now defaults to true. Settings files stamped version 13 by the builds that shipped the block with `enabled: false` are corrected as well: the settings layout version rises to 14 and the upgrade replaces the whole `activity_sdk_generation` block with the bundled one. The block is replaced rather than its member because `enabled` is not unique in the document, and the block replacement also covers the `lua_declarations` default that version 13 turned on.

Changes since the first revision: rebased on current master after the tests folder was removed, so the test file is gone; the port helper stays in `peer_out_of_band.cpp` instead of a new header; the one remaining comment that described world-state 8 as post-spawn is corrected; the parser no longer calls the SDK gate fail-closed; and the settings version bump with its upgrade row was added. Release x64 build is clean.

The connection and spawn changes use native C++ server behavior, and the upgrade correction fixes the native initialization default for required SDK metadata. No gameplay scripts or client-side workarounds are added. Travel-animation changes are outside this PR's scope.

### DIFF
--- a/Sunrise/resources/default_settings.json
+++ b/Sunrise/resources/default_settings.json
@@ -1,5 +1,5 @@
 {
-  "version": 15,
+  "version": 16,
   "core": {
     "logging": {
       "debugger_sink": true,
--- a/Sunrise/src/client/content/activity/activity_sdk_generation_worker.h
+++ b/Sunrise/src/client/content/activity/activity_sdk_generation_worker.h
@@ -62,7 +62,7 @@ struct OfflineBuildResult final {
                                                void* progressContext,
                                                OfflineBuildResult& output) noexcept;
 
-/** Immutable boot policy for the live generator. Each output nothing reads is opt-in. */
+/** Immutable boot policy for the live generator. Its pack backs host roster mission seeds. */
 struct Policy final {
     bool enabled{};
     /** Writes the sdk/lua declaration tree, which no runtime loads. */
--- a/Sunrise/src/core/settings/settings.h
+++ b/Sunrise/src/core/settings/settings.h
@@ -13,10 +13,10 @@
 
 namespace sunrise::core::settings {
 
-/** Boot policy for the optional full-estate activity SDK generator. */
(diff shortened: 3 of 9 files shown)


stanuwu:
reviewed: low code quality
- floating magic numbers

chnsw:
Rebased on current master and reworked after review. The test file with the literal ports and addresses is gone, and the port helper is back in peer_out_of_band.cpp as a local function instead of a new header; the only literal left is the zero-port fallback, which Endpoint already documents as the primary port. Also corrected the last comment that still called world-state 8 post-spawn, fixed the parser doc that called the SDK gate fail-closed, and added the settings version bump (14) with an upgrade row so files stamped 13 with enabled: false are corrected too. Release x64 build is clean.

chnsw:
Verified the upgrade path in game: built the 0.3.2 tag, let it write its own settings.json (version 6, no activity_sdk_generation block), then launched a 0.4.0.0 build carrying this branch over that file. The file came back stamped version 14 with the block still absent, the activity host built its roster, and the Tower loaded to in_world with no errors in the log.

stanuwu:
Merged, thanks.

PR #105, opened 2026-09-05
Fix truncated map-generator Auth records

chnsw:
The server truncates the empty type-37 map-generator Auth record, which can prevent the client from applying the activity roster and leave loading stuck. The encoder counts schemas `0x8080956C` and `0x8080954D` as single bytes, but they are fixed arrays of 32 and 64 bytes.

Correct the empty body from 998 to 1,750 bits so the server sends the complete layout expected by the client. This fixes the shared native C++ encoder while keeping map-generator objects enabled.


stanuwu:
Merged, thanks

PR #98, opened 2026-09-05
Fix squad spawning in the player’s current instance

Confetti3:
Accept multiple individually exact spawner-to-rule references, and make Activity Host selection follow the host for the player's current region. Missing, stale, ambiguous, or unknown ownership still rejects selection. Turning **Follow player** off keeps manual instance selection available.

SDK shard and pack cache identities advance to v14 so previously rejected squad associations are regenerated. Placement activates the authored spawner; selecting a rule does not force enemies onto its displayed anchor points.

Review cleanup: matched nearby upstream Doxygen contracts, clarified publicity lookup results, reused named session/region sentinels, named the unbound generation, and removed the test fixtures and all CMake changes from this PR in line with upstream's shipping-only tree policy. Regression tests run from external local tooling against the revised sources.

Validation:
- Four external CTest targets pass, including squad associations and host ownership regressions.
- Release x64 build: zero warnings/errors. Independent source review and whitespace check passed.
- The original behavior was confirmed in the user's before/after game test. The review revisions refine comments, name existing sentinel values, and remove test infrastructure; no new runtime test or deployment is claimed.

https://github.com/user-attachments/assets/bb64b4f0-52fa-40fe-bfcf-0c18f13cce06

https://github.com/user-attachments/assets/487ae178-5ede-4c9e-9d5d-3ce7bb7ca36d

PR #106, opened 2026-09-06
Keep launched Sunrise tools in a shared workspace

Confetti3:
Sunrise's launched tools currently open as separate floating windows. This keeps the standard main menu and settings layout, but puts World / SDK and Packets in the same workspace so they are easier to find and switch between.

## Workspace changes

- Keep the original animated Sunrise mark as the entry back to the main menu. Launched tools appear beside it as closable, reorderable tabs; opening an existing tool focuses its tab.
- Give each tool the content area, remove redundant window wrappers and headings, and retain each tab's scroll position. Narrow layouts provide scrolling and an overflow list, with enough room for labels and close controls.
- Preserve dragging, resizing, DPI scaling, and double-click maximize/restore. There are no separate minimize/maximize buttons. Reset layout restores geometry without closing tools or changing selection.
- Save geometry, maximized state, tab order, and active selection in `Sunrise/workspace.json`. Migrate older settings and discard unavailable tool IDs safely.
- Register launched-tool content through the existing module registry and keep World and Packets on the same Activity Host selection.

https://github.com/user-attachments/assets/667145c1-5a51-460e-99d9-30b8e802b5e6

## Helper behavior

Enabled World helpers remain visible when switching tabs, returning to the main menu, or hiding the workspace with Insert. Explicitly closing World clears its helpers and published geometry. Cleanup also works when a tool is removed or the viewport temporarily has zero size, without clearing other open tools' helpers.

The marker implementation and header are unchanged from upstream. Existing activity/context validation remains in place; inactive interactive pages are not drawn.

## Included fixes from #98

This also includes the squad-spawning and Follow player fixes from #98: accept multiple individually exact spawner-to-rule references, invalidate stale SDK caches, and select the host for the player's current region without guessing when ownership is missing or ambiguous. Manual instance selection remains available with Follow player disabled.

## Validation

- Release x64 build and all seven focused CTest targets pass.
- Coverage includes tab opening/deduplication/reordering/closing, overflow, persistence and migration, missing tools, shared instance selection, helper retention and cleanup, and squad associations.
- Actual ImGui renders inspected at wide and narrow sizes and multiple DPI scales, including installed-font label clipping and spacing checks.
- `git diff --check` and consolidated independent review pass.
- The final build was installed with verified DLL/PDB hashes and tested in-game by the author before publication.


stanuwu:
No planned
Pop out windows was made for potential multi tasking as well as supporting resize without affecting the main window.

PR #108, opened 2026-09-07
Add a hook to suppress Turn Back

FunTust:
Added a "Turn Back" toggle to the movement panel to disable the out-of-bounds warning and countdown for the local player. The setting is saved between sessions and is off by default.

Moved controlled-object lookup into a shared module used by both Turn Back and teleportation.

Added call-site validation, error reporting, and safer hook removal that preserves state if detaching fails.


FunTust:
I'm sure someone will ask, so I'll clarify this upfront: I developed this implementation independently and didn't use #69 as a reference. I only came across that PR when I was getting ready to submit my own.

PR #109, opened 2026-09-08
Unlock collectibles in the default investment database

fate6:
## Summary
- unlock the tested account-wide cosmetic ownership rows in newly generated investment databases
- seed class emblem objective progress for all three classes
- seed raid guide and Return TYRANT++ objective progress

## Validation
- generated the default database from `investment_schema.sql` and `investment_defaults.sql`
- SQLite `PRAGMA integrity_check` returned `ok`
- built the Release x64 DLL with MSBuild using the 64-bit MSVC compiler
- verified in-game that a fresh database has the tested items unlocked by default

Existing databases are unchanged; these rows apply during new database initialization.


stanuwu:
Merged, thanks

PR #111, opened 2026-09-08
Fix initial_state mission script declaration not propagating to client slice-set streaming and roster publication

ngws:
## Summary
This PR completes the implementation of the `initial_state` feature for mission scripts. Previously, while the VM successfully captured the `initial_state` region and prepared the mission seed on the server side, the declaration was not propagated to the network/transport layer. This resulted in a desync where the server built the logical world for the target region, but the client was instructed to stream the activity's default slice-set (usually region 0).

## The Problem
Because the client was not natively loading the target slice-set, several safety mechanisms in the BAP layer would prevent the roster from being published, leaving the client stuck on a black screen:
1. `arrival_slice_set` ignored the VM's region and fell back to the default destination.
2. The `hostedBubbles` mask did not include the target bubble, causing the server to silently refuse publishing the seed.
3. The `regionArrivalPending` flag was set to `true`, blocking the roster push from reaching the client.

## Changes
* **BAP Layer Override Map:** Added `set_script_initial_slice_set` and `get_script_initial_slice_set` in `bap_route.cpp` to store the script's declared initial slice-set per session.
* **VM Bridge:** Modified `mission_script_runtime_attach.cpp` to populate the BAP override map when `initial_state` is successfully captured.
* **Native Slice-Set Streaming:** Modified `arrival_slice_set` in `activity_arrival.cpp` to check the script override first, ensuring the client natively streams the correct physical map geometry from the start.
* **Dynamic Hosted Bubbles:** Modified `activity_roster_snapshot.cpp` to dynamically include the script's target bubble in the `hostedBubbles` mask, allowing the server to publish the correct roster groups without bypassing the global safety check.
* **Targeted Arrival Bypass:** Modified `select_activity_mission_seed` in `bap_route.cpp` to bypass the `regionArrivalPending` publication blockade *only* when the selection matches the script's initial state. This allows the roster to reach the client immediately without breaking mid-activity transitions.

## Impact
This ensures that missions declaring an `initial_state` (such as `mission_towerfall` loading directly into Underwatch) will natively stream the correct slice-set and receive the world data immediately, bypassing the black screen deadlock without compromising the safety checks required for mid-activity transitions.


stanuwu:
reviewed: some issues
- message 1 always has a default body at first because the first one is sent before the script attaches, need to make the script attach earlier
- difference between public and private link is not respected properly
- state was added in form of a map that is not owned or cleared
- bap runtime is not supposed to be used inside the server, its an outward facing interface
- validation checks are skipped
- doxygen is duplicated in header and cpp, sunrise uses a different convention where only one holds the full block -> check existing code
- some other formatting issues, again check existing code
please fix for merge

ngws:
Thanks @stanuwu good review. In fact, this whole approach was implemented the wrong way round.

I've completely rearchitected the approach. Please review if you're happy with the new implementation, and if so check if you're happy with the formatting and comments.

**Old Approach (Hacks & Back-to-Front)**:

- Data Flow: The server pushed data directly into BAP (the outward-facing protocol layer) using an unowned, global `std::unordered_map`.
- Timing: Relied on the asynchronous polling loop, meaning BAP built Message 1 before the VM attached.
- Safety: Globally bypassed safety checks (`hostedBubbles`, `regionArrivalPending`), which would break mid-activity transitions in other missions.

**Current PR (Native Architecture)**:

- Data Flow: The server writes to the State Layer (`SessionRecord.destination`). BAP simply reads from the State Layer natively, as it was always designed to do.
- Timing: The synchronous `register_instance_active_callback` ensures the VM attaches and writes to the state layer before BAP builds Message 1.
- Safety: Uses the native `sliceSetOverride` mechanism, which inherently respects public/private links, owned memory, and existing validation. Zero safety checks were bypassed.

ngws:
Working on a reimplementation of this since the original implementation was the wrong way round, to be honest.

ngws:
Thanks @stanuwu good review. 

Changes:

The previous implementation was reverted.

Instead, we are now probing the mission script's `initial_state` during session preparation using a lightweight Lua VM, and set `sliceSetOverride` on the `DestinationSelection` _before_ Message 1 is built.

- Implemented `apply_script_initial_state_override` — reads the script source via existing `read_source`, calls the probe, and sets `sliceSetOverride`
- Added `probe_initial_state_region` — executes the script in a minimal Lua state and reads `initial_state.region_index`
- The new `apply_script_initial_state_override` is called before `prepare_session` in `activity_host_manager_route.cpp` and `group_host_sessions.cpp`

An initial attempt that used a callback from the host runtime into the mission runtime was reverted due to lock recursion (the callback fired inside the host's exclusive lock while attach_instance performs heavy I/O).

> Note: This change gracefully handles missing scene seeds in the SDK catalog. Ideally the SDK extraction should produce complete data, but this ensures the server remains functional when coverage is incomplete.

stanuwu:
- g_source (mission_script_runtime_attach.cpp:3) can race because there is no lock for the public entry point, every other one acquired g_lock
!! real fix for above is not the lock, realistically this should not run on the request thread, probably should run on the server thread

- activity lookup still skips checks

- seed change publishes partial roster

- refusals do not debug log

ngws:
ACK, will fix these. 

The only caveat is that the `continue` in `collect_scene_seeds` (or an equivalent) is necessary for this functionality to work at the moment, since the SDK contains incomplete scene resources and can fail the entire roster if we don't include this. The proper fix is to fix the SDK generation. We could skip failed resources in activity_sdk_seeds.cpp -> `materialize_authored_scene_seeds` instead. But we need some way to resolve this. Let me know what you want to do here. I'm testing this with `mission_towerfall`.

stanuwu:
Ofc best way would be to fix sdk generation if possible. If its out of scope just use placeholder values.

ngws:
Updates/changes:

- I have added a more targeted fix for missing resources in `activity_sdk_seeds.cpp` with a warning log, since I verified that SC_UNDERWATCH_INTRO is genuinely missing from mission_towerfall despite holding a slot, so the SDK must respect this and not hard fail mission seeds
- For now, I'm keeping the override application within the request thread because it is nanosecond quick - the alternative would be holding/delaying message 1 server-side until the probe is done, which would take longer and is more complicated
- A small fix for the SDK generation when discovering authored scene linked squads - the current functionality assumes 1 linked squad only, but there can be multiple entries for this table
- Added as many activity checks as is feasible at the point of the probe - we don't even have a session at this point so we can't do the full suite of checks that sdk::resolve would do

If you're still uncomfortable with the request thread triggering the override application then this will need a rework to hold message 1, but I think that's also messy even if it's technically "more correct". Since the Lua mission scripting is not a native game component, any approach we take is inherently unnatural anyway.

The alternative to both of these approaches would be making initial_state arm an immediate teleport instead.

stanuwu:
Sorry but I'm gonna have to close this
Its still not a server based implementation like I already pointed out, it relies on a client hook that is slated for removal which is not a real fix...
"!! real fix for above is not the lock, realistically this should not run on the request thread, probably should run on the server thread"
This was pretty much completely ignored so I will be closing the PR

PR #113, opened 2026-09-12
Fix activity loading on Linux and add a Proton launcher.

itstxl:
This fixes infinite activity loading on affected Linux filesystems and adds a one-command Proton launcher, making Sunrise straightforward to run on Linux.


stanuwu:
not planned
proton launcher should not be needed, the fix would be proper support internally not a workaround

PR #114, opened 2026-09-14
Charge vendor sale rows their own cost (Tower vendors)

mozoii:
Tower vendor purchases spend the price the row names. #87 landed the vendor
interactions with the cost deliberately unspent, on the grounds that the row's
cost-bearing fields were still role-open. This settles that role and charges them.

The role is settled by the data rather than assumed. A sale row's `+32` array holds
up to four 48-byte cost entries (class `0x80807865`). Each entry is an item index at
`+0`, a static quantity at `+4`, two expression-array descriptors at `+8` and `+24`,
and a word at `+40` that is `100000` on every expression-free entry. Of the 5,069
entries across the 4,364 priced rows in the installed catalog, 4,706 carry no
expression and the plain word; every one of Xûr's does, and his rows come out at the
documented Legendary Shard prices: 29 for a weapon, 23 for armour, 97 for the Fated
Engram and 9 for an Invitation of the Nine.

- **The whole cost entry decides whether the static quantity is the price.** A row
  whose entries are all expression-free with the plain word is `plain` and charges
  every entry together. An entry carrying an expression, or another word, makes the
  row `conditional`: its price depends on state the expression evaluator does not
  implement (the programs use `loadValue` and numeric opcodes beyond the five it
  knows), so the row refuses at purchase as `conditional_price` rather than being
  sold at a guessed price or for free.
- **A bad cost array costs the row, not the vendor.** A purchase names a row by
  ordinal, so an unreadable row stays in the bank with its item and category, refuses
  as `unreadable_price`, and is logged once at extraction as
  `ev=build_data stage=vendors result=unreadable_cost hash= row=`. The vendor's other
  rows are unaffected.
- **Paid through the engine that already existed.** `prepare_item_acquisition` and
  `prepare_profile_item_acquisition` take the row's cost entries as an optional price,
  spent in place of the collectible's material set through `apply_material_requirements`:
  it checks each cost item is a stackable profile item, sums every stack, refuses if
  short, then debits and compacts. A Collections pull passes no price and pays as
  before.
- **Repeatable bounty tiles pay their listed price.** The tile prices whatever it
  rolls; only its placeholder item is withheld.
- **Recycle rows require exactly one plain entry**, since an exchange charges one stack.

What the conditional rows are, from dumping every one of them in game (407 entries on 282
rows across 11 vendors): the `+8` program is a per-entry visibility condition
(`loadValue slot`), and the `+24` program computes the price. Banshee-44's legendary mod
rows carry four entries each (mod component, 10 Legendary Shards, 25 of one planetary
material, 5,000 Glimmer), every entry gated by its own value slot, so charging all four
would be wrong. Eververse's 139 Bright Dust rows carry a price program whose constants
read as a computed price (600 against a base of 800 on the sale category, a flat 200 or 333
behind a condition on the rest). Those programs use opcodes the evaluator does not
implement, so both vendors refuse rather than sell at a guessed price; decoding those
opcodes is the follow-up that opens them, and it belongs with the interaction-gate work,
not with the charge.

Refusals stay fail-closed, on the rule #87 set. A cost item the engine cannot debit
or too few units refuses and logs `ev=vendor stage=price result=fail item= quantity= held=`,
and neither grants anything. The charge and the grant are one prepared mutation, so a
purchase that cannot pay leaves the account exactly as it was.

The cache record carries every entry and the row's price state; cache format moves
65 to 66, so an existing `build_data.bin` rebuilds once.

Observable in game: buying from a Tower vendor debits the stack the row names, a
purchase you cannot afford grants nothing, and a row whose price the build cannot
settle refuses instead of being sold free.

Verified in game with Xûr, whose rows exercise every price in the set above:
purchases granted the item and debited the stack, and a row the account could not pay
granted nothing. Rebased onto 0.5.0: both `Release|x64` and `Debug|x64` build at /W4
with warnings-as-errors, and clang-tidy is clean across every file this touches.

Footprint against master: 11 files, +368 / -116, none new.


stanuwu:
reviewed:
- general low code quality/clutter, please integrate better
- price is not read properly, cost row has more content than what is read, it is not confirmed that the method provided always results in the correct price (look at materials for example, they have a condition)
- for invalid value the whole vendor should not be skipped, only the row

mozoii:
No problem. I will get that updated shortly.

mozoii:
Ready for re-review.

mozoii:
Rebased onto 0.5.0 (two commits on 5e4cbc7, cache format 66 vs master's 65).
Release and Debug build clean at /W4, tidy and format clean on every touched file.
0.5.0 touches none of the vendor cost path, so nothing here changed on the rebase.

Since the review:
- state/vendors layer removed; 11 files, +368 / -116, none new.
- The whole 48-byte cost entry is read. An expression at +8 or +24, or a trailing
  word other than 100000, marks the row conditional and it refuses at purchase.
- A bad cost array drops the row, not the vendor.

I dumped every conditional entry in game (407 on 282 rows, 11 vendors) to check
that refusing is right. +8 is a per-entry visibility condition: Banshee's mod rows
list four entries each gated by its own value slot, so charging all four would be
wrong. +24 computes the price: Eververse's Bright Dust rows produce e.g. 600 on a
base of 800. Both use opcodes (12, 14, 17, 18, 19) that Sunrise's expression
interpreter does not implement yet, so their price cannot be determined. Those
rows refuse to sell rather than charge an unconfirmed amount.

PR #115, opened 2026-09-14
Add player-hosted co-op for shared exploration

Techno453:
Adds support for separate Sunrise installations to form fireteams and explore shared activities.
One player's game hosts the shared session services inside its DLL; guests connect to that host.
Each installation keeps its own account database, identity and SDK output. Closing the hosting
game ends the shared session. Multiplayer requires explicit opt-in and defaults off.

Incorporates upstream 0.5.0, preserving its activity activation controls, mission
destinations and arrival spawn selection. No client changes were added after the 0.5 merge.

## Account ownership and social features

Private account reads and mutations stay on the owning installation. Guests forward enrollment,
public profiles, presence, invitations, chat, fireteams and activity membership to the host.
Request scopes select the appropriate account; unresolved identities cannot fall back to the
host's private account. Local, request-scoped and public snapshot APIs have separate contracts.
SQLite shutdown follows process ownership independently of the current request scope.

Roster and Inspect use the target player's published character, equipment, Power and emblem.
Remote artifact Power comes from published presence. Public encoders omit private inventory
mutation counters and progression banks.

The friends interface receives session members through the platform adapter. Persona callbacks
wait for the native friends manager and account registration. Invitations use an Accept/Decline
overlay; delivery alone does not join the player. The callback pump owns invitation progression;
the UI reads a coherent snapshot and posts ID-tagged decisions without waiting on a lock.
Pending delivery recovers when inbox or callback capacity returns. Fireteam chat uses platform
lobby records; membership follows committed native reservations.

## Shared activities and transport

Reservations, membership revisions, entity leases and published endpoints track players through
departure and return. Native clients continue to produce, serialize and apply gameplay replication.

- Include the playing host's published endpoint in relay authorization, with the same exact
  source/destination checks as other peers. Quiet healthy routes survive without periodic social
  refresh requests; authenticated withdrawal and connection lifecycle remove stale ownership.
- Publish top-level roster records through their private owning connection and preserve the
  original owner's native reservation for a subsequent return.
- Keep replication epochs shared across the activity session and deliver them consistently
  through join results and member departures. Release an account's owned group admissions
  when its final authenticated connection closes.
- Resolve mission-program ownership from members still present in the native activity.
- Withdraw presence at character selection and publish roster changes when players leave
  or change characters.
- Keep committed reservation identities visible when reconnect changes the selected character.
  Optional name metadata cannot hide a reservation; native release precedes fresh admission.
  Transport still has to match the reserved account, character and member key.
- Derive arrival from committed ActivityClient region reports, not the WS-702 join-controls
  field. Only explicit checkpoints establish persistent respawn overrides.

Co-op uses one configurable port number, 30974 by default, for TCP control and UDP discovery,
activity channels and peer traffic. The Winsock adapter adds a 16-byte logical-endpoint header
to native UDP payloads. The host dispatches them to a service or relays them within admitted peer
pairs; the receiver restores the logical source before returning the native payload to the game.
Carrier datagrams are serviced on arrival with bounded work per service call.

Host TCP connections use keepalive after 30 seconds idle, followed by ten one-second unanswered
probes. Guest BAP keepalive deadlines pause under local output pressure. Route cleanup continues
when no local host BAP links remain, preserving live guest routes and withdrawing stale owners.
Added `mstcpip.h` to the system-header whitelist for `tcp_keepalive` and `SIO_KEEPALIVE_VALS`.

Discovery uses the game's socket so the host observes its mapping. A single physical endpoint
cannot perform the independently sourced NAT probes correctly: unsupported stages remain
unanswered and the native client conservatively falls back to strict NAT. Logical port changes
are not treated as proof of open NAT. The host must be reachable on the selected TCP and UDP port.
Automatic router forwarding is not implemented; relaying uses the host's bandwidth. The supported
carrier uses synchronous `sendto`/`recvfrom`; alternate UDP operations remain unsupported.

## Request delivery and client integration

Ordinary social/profile publication follows client requests, real content/dependency changes
and output availability. Explicit subscriptions and WS-206 family fetches receive snapshots;
arbitrary frame-count replay is removed. Retained clocks serve transport retry, liveness or
resource cleanup. Bounded proxy queues preserve input/output, correlations and nonces under
backpressure. Native forwarded requests retain reply order; notifications can drain independently.
Established mutations are not replayed after reconnect. Request paths reuse bounded storage;
shared state follows SRW locking and documented ownership conventions.

Malformed social notices and rejected public-profile projections fail explicitly. Explicit WS-206
fetches retain subscriptions until unsubscribe or disconnect and count toward the 32-root family
limit; refused fetches receive correlated replies.

The reliable-request admission and replication-budget hooks are removed. The presence adapter
is retained because native member-row rebuilding can omit the publication marker: changed output
requests the existing native urgent publisher only for a stable account, character and cache
lifetime. Unchanged output and already-pending publication do not request another publication.
Native serialization and timing remain owned by that publisher.

Retained reviewed hooks use unique patterns and decoded references, refusing missing or ambiguous
matches. The obsolete `composition_check`, `owner_activity_slot` and `region_private` hooks and
their settings/wiring are removed. Direct-socket selection, configured machine-ID override and
local multiple-instance support remain. Comments explain ownership, threading, failure/retry
behavior and protocol/capacity constraints.

## Configuration and compatibility

Host configuration in `bin/x64/Sunrise/settings.json`:

```json
{
  "multiplayer_enabled": true,
  "host": "192.168.1.50:30974",
  "steam": { "user": { "persona_name": "Host" } }
}
```

Replace the address with the host's reachable IPv4 address. Guests use `server` instead of `host`
and their own name; they also need `multiplayer_enabled: true`. Omitting the port uses 30974.
Without explicit opt-in, multiplayer traffic remains blocked and upstream local solo behavior
is retained. Opt-in permits configured multiplayer paths, not unrestricted egress.

Both sides need the same build. Settings version is 19; public-profile wire version is 6.
Older full/versioned settings follow upstream's replacement policy and reset to bundled defaults,
including multiplayer disabled. Unversioned compact settings are preserved. Compact host/client
settings generate a cached identity without an explicit Steam ID; an explicit ID does not overwrite
that cache. Seed databases are rebound once; databases already bound to another identity refuse
startup.

## Validation

Release and Debug builds and automated tests pass. Final-scope clang-tidy reports zero
findings. The final production DLL passed attended solo Moon movement/death-respawn
and two-player join, leave/rejoin, character-change roster and revive smoke tests.
Temporary incident diagnostics are removed from the final tree.

## Limits

Enrollment and relay binding assume trusted participants and do not provide credential-backed
player authentication. Mission/enemy replication remains separate. Research, tests and capture
tooling stay private and outside this PR.

Known issue: the death notification name can be blank for players who joined after
the observing client. The fix is deferred to a future PR because of its scope.


stanuwu:
first quick review:
- hooks use fixed offsets either entirely or for compate, this should use only a unique pattern scan like all the other ones do
- reliable_request_admission is a client fix for a server bug
- presence_publication and replication_budget need a reason to exist (are they required and why)
- nat probe stage 2 skipped (whats the reason)
- try to avoid requests sent based on timers or counters, a lot of that here and its better to respond based on what the client sends instead, should work for the most case just responding based on whats sent in a stateless way
- ws-702 world-state field likely incorrect
- composition_check, owner_activity_slot and region_private hooks were all removed upstream (this will be pushed soon), try to avoid them
- make multiplayer functionality being enabled at all an explicit setting for now (egress guard should block fully unless disabled by choice)
- account scoping defaults to host?, account_snapshot() is a bit ambiguous now, maybe better naming or split, shutdown skips sqlite close in some cases
- proxy queue 8 or 4 drops without retry
- pending invitation stuck in a lock... i dont think we should lock on the client side at all
- settings version needs a bump
- inconsistent use of std::mutex vs SRW lock, also dynamic heap allocations on requests can be dangerous
- quite a few floating numbers or hardcoded values that are not justified/explained
- broadly missing doxygen and some missing comments, some comments are way too much

Techno453:
1. **Fixed hook addresses:** Replaced with unique pattern scans; missing or ambiguous matches
   refuse installation.

2. **`reliable_request_admission`:** Removed the hook and addressed server replies, relay
   scheduling and reply ordering. Unsupported parameters receive released-mask answers, but
   those do not clear the native client's pending bit. Unconditional retransmission can still
   rebuild the backlog.

3. **`presence_publication` / `replication_budget`:** Removed `replication_budget`.
   Retained `presence_publication`: native member-row changes can omit the publication marker,
   delaying roster updates. It requests the existing urgent publisher only for changed content
   within a stable owner/character/cache lifetime, with no publication already pending.

4. **NAT classification:** Address discovery works. Classification remains unsupported because
   the single endpoint cannot supply the independent response the probe requires; native
   strict-NAT fallback remains.

5. **Timer/counter-driven requests:** Replaced periodic polling, publication gates, BAP-frame-count
   replay and fixed rejoin resends with request/content/capacity-driven delivery. Remaining
   clocks handle transport retry, liveness and cleanup; correlations and acknowledgements still
   require state. 

   Hosts use TCP keepalive: 30 seconds idle, then ten one-second unanswered probes. Guest BAP
   keepalive deadlines pause under local output pressure. Route cleanup works without local
   host links. Added `mstcpip.h` to the system-header whitelist for `tcp_keepalive` and
   `SIO_KEEPALIVE_VALS`.

   Explicit WS-206 fetches retain subscriptions until unsubscribe/disconnect, counting toward
   the 32-root family limit; refusals receive correlated replies.

6. **WS-702 field:** Corrected to native join controls; arrival comes from committed region reports.

7. **Retired hooks:** Removed all three hooks and their wiring/settings. Restored upstream solo
   membership behavior. `composition_check` and `owner_activity_slot` were solo-only and enabled
   by default.

8. **Explicit multiplayer setting:** Added default-off `multiplayer_enabled`; multiplayer egress
   requires opt-in. Friend announcements and invitations intentionally remain multiplayer-only.

9. **Account scope and shutdown:** Separated local/request/public access, blocked private-account
   fallback, and made SQLite teardown independent of request scope.

10. **Proxy queue drops:** Added bounded backpressure with preserved reply ordering; malformed
    notices and rejected projections fail explicitly.

11. **Pending invitation lock:** Moved progression to the callback-pump owner; UI decisions no
    longer wait on invitation/server locks.

12. **Settings version:** Bumped to 19 and public-profile wire format to 6. Older versioned files
    follow upstream replacement semantics. Unversioned compact files and cached identities are
    preserved; missing multiplayer opt-in is diagnosed and refused, without migration.

13. **Locks and allocations:** Adopted SRW conventions and reusable storage. Startup, connection
    and activity-lifetime allocations remain. Activity storage still reuses inactive slots first;
    allocation refusal preserves queued input but does not undo earlier lifecycle cleanup.
    The group-release lock still spans State calls; no re-entrant path was found.

14. **Unexplained values:** Documented provenance/derivations and consolidated constants.
    Chosen deadlines and capacities remain policy choices.

15. **Doxygen/comments:** Corrected API contracts and removed stale or redundant comments.

Techno453:
Adopted 0.5 changes. Improves native co-op session handling after 0.5: fixes replication epoch delivery, mission-owner selection, and roster/presence cleanup when players leave or change characters. No client changes since the 0.5 merge.

Validated with Release/Debug builds and tests, both solo and 2 player coop tested.

Known issue: Blank death line for players joined after the specific client. Fix is large enough in scope such that it will be easier to add in a future PR.

PR #119, opened 2026-09-16
SDK: record unresourced authored scenes and read every scene participant

Fozkais:
## Summary

- records, in the SDK pack, an authored scene whose package config references no resource, instead of dropping its resource row and leaving the pack unable to tell it from an extraction failure.
- keeps a distinction the seed cannot carry: `materialize_authored_scene_seeds` writes the absent resource tag for such a slot, which reads the same as a resource that was referenced and could not be read. The flag separates the two, and a referenced resource that is unreadable or of the wrong class is logged as an extraction defect.
- reads the complete participant table of an authored scene and emits one scene-to-squad edge per squad participant; the generator read a single block at the position it has when the table holds exactly one entry.
- names the target squad in an edge's identity and natural key, since a scene may carry several.

---

- Adds `kSlotAuthoredSceneUnresourced` to the slot flags and the slot flag mask.
- Replaces the two fixed squad-block offsets with the participant table layout: count, table class, self-relative pointers, block class before the payload, squad reference inside the payload, and a participant bound.

##### `Sunrise/src/client/content/activity/activity_sdk_authored_scene_inventory.h`

- Adds the list of scene slots whose config references no resource to the inventory snapshot.

##### `Sunrise/src/client/content/activity/activity_sdk_authored_scene_inventory.cpp`

- Records a scene whose resource field is absent; logs a referenced resource that is unreadable or of the wrong class as an extraction defect.
- Walks the participant table and emits one edge per squad block, bounded by the blob and the participant capacity.
- Includes the squad slot in the edge identity.

##### `Sunrise/src/client/content/activity/activity_sdk_authored_scene_internal.h`

- Extends the edge natural key and identity signature with the squad slot.

##### `Sunrise/src/client/content/activity/activity_sdk_authored_scene_validation.cpp`

- Validates the unresourced list: ordered, unique, scene-shaped, and never also resourced.
- Bounds a scene edge's reference by the participant table's position rather than a fixed offset; performance edges keep theirs.

- Adds the pipeline step that carries the unresourced decision onto the slot rows, after the scene inventory and before the dialogue cue counts.

##### `Sunrise/src/client/content/activity/activity_sdk_lua_contract.cpp`

- Exports `unresourced = true` on such a slot and declares the field in the Lua contract.

---

## Validation

- Release x64 build of `535b351` passed.
- clang-format passed on all 10 affected sources and headers; the configured clang-tidy passed on all 6 affected translation units.
- Pack regenerated from the installed content. Across it, 50 scene slots in 10 packages reference no resource; every referenced resource resolved (no `unreadable` or `class` result).
- Scene-to-squad edges across the installed content went from 682 to 2231. For scenes read directly from their package configs, the emitted edges match the squad blocks of the participant table, including scenes with three and four participants.
- Performance (type-42) edges and the generated `PerformanceState` tables are unchanged.

## Notes

- A scene authored without a resource remains a component on the client; nothing is seeded for it, and a script cannot bind it.
- This branch does not touch the seed path: `materialize_authored_scene_seeds` writes the absent resource tag for such a slot, as it does today.
- The edge identity string gains the target squad's slot index and type; the version bump covers existing packs.
- The Activity Host's SDK view shows an edge count per scene, not the linked squads.


Fozkais:

Fozkais:
Update: the fix from [033284c](https://github.com/stanuwu/Sunrise/pull/119/commits/033284c609aeacf5605cf4f6cbd5185216185d26) causes issue, game seem to have a limit of 8 dependencies for a scene, but some scene have 12+ participant, so this wont work, i'm investigating how we can determine wich participant is really needed as a dependencies so i ajust the way it work. Other part still apply.

Fozkais:
Update: fixed in 06e3b6b. I reversed the client's scene sensor to see what it actually does with those references, and it turns out they are not the participant list at all: the client binds every role itself from the participant table in its own content, so a scene plays fine without them. The only thing it does with the references is, once the scene is running, mark each named squad's remaining spawn budget as consumed (the squad won't keep spawning members the scene has taken). The 8 cap is a hard schema limit, and the installed content even has scenes with 9 squads that all have actors, so Bungie's server couldn't send every participant either.

So the pack still reads every participant (that part of 033284c stays), but the server now only sends as dependencies the squads it actually casts, i.e. the ones it supplies an actor for through a type-2 cell. Participants without a cell are simply left to the client, which resolves them by itself, instead of refusing the whole scene. If a cast is still wider than 8, the server sends no dependencies and logs `ev=scene_dependencies result=omitted`, which is what 0.5.0 did for every multi-participant scene.

PR #122, opened 2026-09-17
SDK: record scene event keys and counter directive elements, carry directive progress

Fozkais:
## Summary

- records, in the SDK pack, the event gates of every authored scene's graph: one row per gate, carrying the event key the server publishes with `set_scene_events` and the gate's ordinal in the scene's progression. A graph that cannot be read or does not have the expected shape leaves the scene without keys and is logged, never refused.
- records the directive elements authored with a progress label and no description. The generator required a description and dropped the counter variant of a directive; an element is now kept when it has a title and either text, and its progress label, source tags and authored flag word travel with the row.
- carries progress values in the directive lane `set_directive` publishes. The type-68 encoder wrote the four signed lane values as zero, so a counter could only show 0 / 0.
- exposes all of it to Lua: `event_keys` on each scene symbol, `progress` and `counter` on each `mission.Directive` entry, and `progress = { current, maximum }` on `set_directive`. `event_keys` is the authored set `scene:send_event{ key }` accepts for that scene; the call itself only bounds the key.

---

- Adds the `authoredSceneEventKeys` section at the end of the section index, with its row `AuthoredSceneEventKey` (scene slot, resource tag, graph tag, gate offset, ordinal, key, flags), and the graph layout: the graph field of a resource, the typed array marker, the gate array and element classes, the element size, the key and ordinal offsets, and a gate bound.
- Adds the packed directive table layout (table, entry and element classes, sizes and field offsets) and the counter bit. `DirectiveElement` gains `progress`, its source tags and `flags`.

##### `Sunrise/src/state/activity_sdk/generation/activity_sdk_pack_writer.cpp`

- Register the event key row stride on both sides and carry the section through the writer tables.

##### `Sunrise/src/state/activity_sdk/activity_sdk_catalog.cpp`

- Add `Catalog::authored_scene_event_keys()` and `slot_authored_scene_event_keys()`, which returns one contiguous range per type-43 slot and relies on the section's slot ordering, as the resource accessor does.

##### `Sunrise/src/state/activity_sdk/activity_sdk_structure_validation.cpp`

- Add the `authored_scene_event_keys` check: every row names a type-43 slot that holds exactly one resource of the stated tag, flags are exact, key and graph are set, and rows come in slot then gate order without repeats.
- Relax the directive checks: a description or a progress label is required, not both; an authored field has both source tags and an absent one has neither.

##### `Sunrise/src/client/content/activity/activity_sdk_authored_scene_validation.cpp`

- Follow a resource to its graph, require exactly one typed array of gate elements, check that every element opens with the graph's own tag and the gate class, and record its key and ordinal. A graph that is unreferenced, unreadable, without a gate array, over the gate bound, or with a malformed gate is logged and yields no keys; only a topology inconsistency fails the build.
- Validate the rows before composition: each belongs to a resourced scene with the same resource tag, its identity recomputes to the stored one, and the order is strict.

##### `Sunrise/src/client/content/activity/activity_sdk_native_pack_text.cpp`

- Reads the three localized references and the flag word of a packed element with the named layout instead of literal offsets; a reference is checked only when present. Keeps an element with a title and a description or a progress label, and logs the ones it leaves out at debug.

- Link and convert the new rows and fields, bound the section and pass it to the Lua source, as for the other authored scene sections.

##### `Sunrise/src/client/content/activity/activity_sdk_lua_missions.cpp`

- Publish `event_keys = { ... }` in gate order on each scene symbol and the section count in the manifest.
- Publish `progress` and `counter` on a `mission.Directive` entry when authored; a counter element takes its key from its progress label, since it shares its title with the described one. Declares `SunriseDirective` and the `progress` argument in the Lua contract and the new directive fields in the catalog contract.

##### `Sunrise/src/middleware/bap/activity_message/activity_scriptable_auth_mission_body_codec.cpp`

- `Type68Preset` carries the four signed lane values; the encoder writes them as given for the active lane and zero for the inactive ones. The body keeps its shape. `write_directive_entry` takes them after the navpoint name and bubble hashes and the waypoint reference.

##### `Sunrise/src/server/activity/mission/mission_script_lua_presentation_api.cpp`

- The resolved element definition carries the authored flag word. `set_directive` accepts `progress`, one to four int32 values, beside the other named arguments, and refuses it on an element that declares no counter, since the HUD would ignore the values.

##### `Sunrise/src/server/ui/activity_host/activity_host_sdk_mission_view.cpp`

- Shows the progress label and marks counter elements.

---

## Validation

- Release x64 build of `e581d0b` passed.
- clang-format passed on all 29 affected sources and headers; the configured clang-tidy passed on all 20 affected translation units.
- Pack regenerated from the installed content: 5310 event key rows; every scene with a resource received its keys and no `activity_sdk_scene_graph` result was logged. For scenes whose graphs were decoded by hand from their packages, the exported keys match in count, value and order, including the shared opening gate (ordinal -1) and scenes with 1, 3, 5 and 10 gates.
- Directive elements went from 14932 to 15076. The 144 added rows are exactly the rows with an empty description, and every one of them carries a progress label; every element the previous reader kept is still present.
- Live, on a counter element with no description: `set_directive{ directive = ..., progress = { 0, 3 } }` shows the element's title once with "<label> 0 / 3"; re-sending it with `{ 1, 3 }` updates the counter in place without a new banner.

## Notes

- Independent of #119 and #123; all three move the pack version from 40 to 41, so whichever merges second rebases and takes 42.
- A scene authored without a resource has no graph to read; it receives no keys and its seeding is unchanged.
- Only the key and its ordinal of a gate are exported; the gate element carries no participant reference and no name.
- The authored flag word of a directive element is carried unchanged; only the counter bit is interpreted. Bits `0x2` and `0x8` occur in the installed content and are not named.
- `set_directive` fills one active lane, as before; the three-lane body is not exposed further.


Fozkais:

Issue #120, opened 2026-09-16
Abilities broken on Void Warlock


Git-SuperNova:
I've noticed that on Void Warlock, after changing any ability on the subclass, that the abilities including glide don't work and their HUD icons are completely hollow. However, when leaving the subclass at default with Scatter Grenade etc, it works totally normally. I have not tested Void on any other classes. I have tried changing the subclass choices in Sundial before launching the game but that doesn't fix it. All other Warlock subclasses can change abilities with no problem. This problem is occuring with a clean settings.json, sdk and cache.

PR #125, opened 2026-09-17
Size the ability row table to every subclass's remembered picks

Yakariii123:
Changing the ability picks of a second subclass on the same character leaves that subclass with no abilities: the HUD draws no ability icons and the movement ability does nothing, while the default picks keep working. Reported in #120 against Void Warlock, though the subclass itself is incidental.

Ability picks are stored per subclass item, so every owned subclass remembers its own, and `build_character_abilities` publishes a row for each: a default-selection row per owned subclass, plus a second row for every subclass whose remembered picks differ. `abilities::kDefinitionCapacity` was still sized for the earlier per-character model, documented as "3 characters by up to 4 rows", so the next row a character needed was dropped by the `count >= output.size()` guard and nothing published the buckets for that selection. The default row survives because it is published first, which is what makes the failure look specific to one subclass. The guard also returned without a report, so the pass still logged `ev=pkg stage=abilities result=ok` while dropping a row.

The table is now sized to what the publisher can need, 3 characters by 3 owned subclasses by 2 rows, and the overflow reports `reason=capacity` with the character and the socket-entry list index. Rows are keyed by subclass and selection and de-duplicated on publication, so that bound holds.

The build-data cache version is bumped with it, so existing installations rebuild. This one is load-bearing: a cache written at the wider capacity fails `counts_fit` on a build that still has the narrow one, which is `LoadStatus::invalid` rather than `stale`, so it fails the boot at the `state` stage instead of taking the rebuild path from #77. Note that #114 also takes 66, so whichever lands second needs the next number.

Reproduced on master with several subclasses changed across three characters; a temporary report at the guard confirmed the cause directly, `ev=pkg stage=ability_failure reason=capacity character=2 first=10 second=12`. After the change there are no `ability_failure` lines, published rows reach 16, and ability icons, abilities and the movement ability work on every subclass of all three characters. Installing a narrow-capacity build over a cache written by this one reproduces the `stage=state` boot failure without the version bump and boots normally with it. Release x64 built with the v145 toolset and the 64-bit compiler host. clang-format clean, clang-tidy clean on the changed translation units.


Yakariii123:
Still applies to 0.5.0. On the shipped 0.5.0 release DLL, with picks changed on several subclasses across characters, the published rows stay at the 12-row cap through every extraction pass (ev=pkg stage=abilities result=ok rows=12) and the affected subclass draws no ability icons and has no movement ability.

With this change on top of 5e4cbc7 the same account state publishes 16 rows, with no ability_failure line, and every subclass works. Since picks persist across sessions, once a character has changed the picks of a second subclass the incomplete set is published on every boot. The branch rebases onto 5e4cbc7 without conflicts.

PR #130, opened 2026-09-19
Evaluate type-30 player monitors on the Activity Host from channel-3 player positions

Dheeraj-Banala:
## What

The Activity Host now measures type-30 player-occupancy monitors itself and raises the same
`triggerEntered` / `triggerExited` events a mission script already handles. It gets the player's
position from external channel 3, which every client already sends and which Sunrise used to skip.

## Why

Type-30 monitors (e.g. Leviathan's Castellum destination plates `pm_flag_destination[0..3]`) never
fire in a private activity. On a client, the monitor counts only players whose owner equals the owner of
the current activity context, which is the Activity Host the client joined. In a private activity that is
never true, by design. The client is not supposed to count these players; the Activity Host is.

The host has what it needs. Channel 3 of the external frame (reflected root `0x80806AE6`) is the sending
client's own player snapshot. Its first data field is the player's world position (3 × f32). A client
in a private activity sends it on every in-world frame (about 10/s). Sunrise read the lane's presence bit
and skipped the 1373-bit body.

## How

1. **Decode channel 3.** `player_snapshot_codec` reads the snapshot in reflected field order (player key,
   position, two u64s, three biased i32s, five bools, i16[64]). It consumes exactly the bits the old skip did
   (`static_assert` on 1373), so frame parsing is unchanged. `read_external` keeps the decoded lane.
2. **Report it.** After an accepted external frame whose channel 3 is present, `consume_established` calls
   `mission::report_player_position(binding, activityClientGeneration, key, position)` once the peer lock is
   released. This follows `report_squad_provoked`: the runtime finds the instance by binding and checks the
   source generation.
3. **Resolve monitors.** On first use, each instance collects every exact type-30 slot (component
   `0x8080952F`, Sense `0x80809531`) that the scriptables catalog already joins to exactly one type-60 trigger
   volume (`triggerVolumeIncomingReferences`, which upstream already builds for types 31 and 30). Ambiguous
   joins are left to the client.
4. **Measure.** `trigger_volume::contains` uses the model upstream's own renderer draws
   (`package_trigger_volume_geometry`): the point lies over a triangle of the floor footprint, between that
   triangle's interpolated floor and floor + `extrusion`. Inactive or incomplete instances contain nothing.
   Up to 6 player positions are kept per instance; the origin (sent before the player is placed) is ignored,
   and a player unheard for 5 s is dropped.
5. **Raise edges through the existing path.** `TriggerOccupancy` now keeps the client's level and the host's
   level separately; the volume is occupied when either is, and an edge fires only when that combined level
   changes, so the two sources never double-fire or flap against each other. The client Sense path keeps its
   exact upstream behaviour (a first observation only records the baseline). The host takes a row only once
   someone is inside, and frees it again when empty, so host evaluation cannot exhaust the 32 shared rows.
   Events carry `triggerCount`, `triggerAll` and the monitor's caller value from its latest type-30 Auth.
6. **Filters.** A monitor that a script gives an object filter (`set_occupancy_condition{filter=…}`) is left to
   the client and logged as `host_occupancy result=filtered`. Evaluating filters needs type-34 predicates on the
   host, which is out of scope. No shipped or authored monitor in Leviathan carries a filter (all 16 observed
   monitors report the absent reference `0x811C9DC5`), and no script calls the API today.


stanuwu:
Confirm against D1 stack and D2 Binary and potentially move to Phys Host Shim

PR #132, opened 2026-09-19
Add configurable Jump Height modifier

saulob:
## Summary

Adds a configurable Jump Height modifier as an additional movement/exploration option.

The implementation extends the existing movement/Havok handling and only modifies the player's vertical velocity.

## Changes

- Added a separate Jump Height toggle
- Added an integer multiplier from 1x to 10x
- Added persistence for the new settings
- Only the initial ground jump is multiplied
- Horizontal X/Y velocity remains unchanged
- Added Guardian component/body lifecycle handling for activity transitions and respawns
- Added narrow Hunter aerial-jump compatibility handling
- Titan Lift and Warlock Glide behavior remain unchanged
- No additional Detours or memory patches were introduced

## Behavior

Jump Height:

- Disabled by default
- Multiplies only the initial ground jump
- Supports values from 1x to 10x
- Aerial abilities are not multiplied
- Automatically rearms after landing
- Continues working across activity/destination transitions and respawns

Hunter aerial jumps include a compatibility adjustment for cases where the modified primary jump can interfere with the native aerial jump behavior. The adjustment intentionally remains narrow and does not multiply aerial jumps.

## Testing

Tested in-game with all three Guardian classes:

- Hunter
- Titan
- Warlock

Tested with multiple Jump Height multipliers and across different destinations.

Also tested:

- Activity/destination transitions
- The Tower
- Landing and repeated jumps
- Titan Lift
- Warlock Glide
- Hunter aerial jumps

Jump Height remained functional across the tested classes and destinations.

## Screenshot

Screenshot showing the new Jump Height option in the Movement menu.


stanuwu:
not planned

PR #133, opened 2026-09-20
Add controller support to Fly Movement

saulob:
## Summary

Adds controller movement support to the existing Fly feature

Fly can now use the controller's left stick for analog movement while preserving the existing keyboard movement controls

The implementation reads Destiny's processed movement vector rather than polling a specific controller API directly

## Changes

- Added controller left-stick movement to Fly
- Preserves analog stick magnitude for variable movement speed
- Keeps movement relative to the current camera direction
- Right stick continues using Destiny's normal camera / aim handling
- Preserves the existing WASD keyboard movement
- Fly and Noclip continue sharing the existing movement path
- Added safe fallback to keyboard-only movement if the controller input signature cannot be resolved
- No XInput polling, Steam Input polling or additional Detours were introduced

## Behavior

Controller:

- Left stick controls Fly movement
- Partial stick input produces proportional movement speed
- Full stick input reaches the configured Fly speed
- Movement follows the current camera orientation
- Right stick continues controlling camera / aim normally
- Diagonal movement is capped to the normal maximum movement magnitude

Keyboard:

- Existing WASD movement remains unchanged
- Existing vertical movement remains unchanged
- Keyboard and controller can be used together without exceeding the configured movement speed

## Controller support

The implementation reads the processed movement vector produced after Destiny's controller backend layer rather than reading Xbox/XInput directly

This should allow other controller types supported by Destiny to use the same path

Live testing was performed with an Xbox Wireless Controller only, so other controller types have not been personally tested

## Testing

Tested in-game in the EDZ, Tower and Titan with:

- Xbox Wireless Controller
- Keyboard WASD movement
- Partial and full analog stick movement
- Diagonal movement
- Camera-relative movement using the right stick
- Fly
- Fly + Noclip

Both controller and keyboard movement worked correctly

The original WASD Fly behavior remains functional

## Screenshot

In-game testing in the EDZ using Fly + Noclip with an Xbox Wireless Controller


stanuwu:
not planned

PR #135, opened 2026-09-20
Add configurable Movement Speed

saulob:
## Summary

Adds a configurable Movement Speed option to the Movement tab

The feature applies a fixed horizontal movement speed in units/s using the existing Sunrise movement infrastructure

It does not multiply the existing physical velocity

## Features

- Configurable horizontal speed from 1 to 100 units/s
- Optional toggle key
- Keyboard movement support
- Xbox controller support
- Preserves analog stick magnitude
- Camera-relative movement
- Normalized diagonal movement
- Works while jumping and falling
- Works together with Noclip
- Fly keeps its existing behavior and takes precedence when enabled
- Sparrow movement speed is not affected

## Implementation

Movement Speed computes the desired horizontal movement direction from the current input and applies the configured speed directly to X/Y

Vertical movement is left unchanged, so Destiny continues handling gravity, jumping and falling

Controller movement uses Destiny processed movement input rather than raw XInput

Fly remains keyboard-only in this PR

## Noclip

Movement Speed can be used together with Noclip

The configured horizontal speed is preserved while Noclip handles collision behavior

Fly + Noclip behavior remains unchanged

## Testing

Runtime tested with

- EDZ
- Io
- Titan destination
- Tower
- Hunter
- Titan
- Warlock
- Keyboard movement
- Xbox controller
- Partial analog stick movement
- Full analog stick movement
- Diagonal movement
- Jumping while moving
- Falling while moving
- Slopes and terrain
- Movement Speed + Noclip
- Fly + Noclip
- Fly and Movement Speed enabled together
- Toggle key
- Enabling and disabling while moving
- Sparrow movement

Sparrow speed remained unchanged during testing

## Notes

Movement Speed also affects horizontal movement while airborne, while Destiny continues controlling vertical movement

At very high configured speeds, running directly into walls can still cause collision damage or death

The player reacquisition change in `player_position.cpp` is the same Guardian reacquisition fix currently present in #132

Movement Speed needs this fix across world transitions, including the Tower

If #132 merges first, this duplicate hunk can be removed after updating this branch from master

## Related PRs

- #132 Jump Height

# Screenshot


stanuwu:
not planned

